PR #17, opened 2018-07-27
py27: change ModuleNotFoundError to ImportError

msarahan:


msarahan:
More details on ModuleNotFoundError: https://bugs.python.org/issue15767

It is a subclass of ImportError.  It was introduced in Python 3.6.  Catching ImportError will catch ModuleNotFoundError too.

delta003:
@oleksandr-pavlyk Can you guys cut a release with this change?

oleksandr-pavlyk:
@delta003 Yes, I am working on it now. Should be out shortly.

delta003:
Thanks!

PR #38, opened 2019-04-16
Bugfix/remove axes

### LINKED ISSUE #37: mkl_fft._pydfti.rfftn_numpy error when using axes specification
I originally posted this as a numpy issue, but after some research it appears to be something wrong with mkl_fft, although it only happens under Python 3 (and only in later releases, where "later" means past whatever the default was in Anaconda 5.0.1).

You can see some background on the issue [here](https://github.com/numpy/numpy/issues/13318) - the context is that I use numpy's rfftn to make a PSD of simulated ocean waves.

Long story short, when I make a conda environment like this:
```
> conda create --name test1 anaconda=2019.03 python=2
```
and this:
```
> conda create --name test2 anaconda=2019.03 python=3
```

If I have a numpy array called 'waves' that is dimensioned [64,512,512] and try to run the following:
```python
import numpy.fft as nfft
realgood2 = nfft.rfftn(waves, axes=[2,1,0])
```
In test environment 1 it works perfectly. In test environment 2, I get this:
```
Traceback (most recent call last):
  File "gist.py", line 57, in <module>
    realgood2 = nfft.rfftn(waves, axes=[2,1,0])
  File "C:\Users\asmith\AppData\Local\Continuum\anaconda3\envs\3_2019.03\lib\site-packages\mkl_fft\_numpy_fft.py", line 1043, in rfftn
    output = mkl_fft.rfftn_numpy(a, s, axes)
  File "mkl_fft\_pydfti.pyx", line 951, in mkl_fft._pydfti.rfftn_numpy
ValueError: could not broadcast input array from shape (512,64) into shape (512,512)
```

oleksandr-pavlyk:
Closes #37.

As noted by @Jeitan, the logic in `_remove_axis` was incorrect. 

Now:

```
(b_mkl_fft) [13:28:08 fxsatlin03 mkl_fft]$ ipython
Python 3.6.8 |Intel Corporation| (default, Jan 16 2019, 05:23:57)
Type 'copyright', 'credits' or 'license' for more information
IPython 6.3.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import numpy as np, mkl_fft

In [2]: x = np.random.randn(64, 512, 512)

In [3]: %time z = np.fft.fftpack.rfftn(x, axes=[2, 1, 0])
CPU times: user 3.98 s, sys: 724 ms, total: 4.7 s
Wall time: 1.12 s

In [4]: %time w = mkl_fft.rfftn_numpy(x, axes=[2, 1, 0])
CPU times: user 660 ms, sys: 92 ms, total: 752 ms
Wall time: 47.3 ms

In [5]: np.allclose(w, z)
Out[5]: True
```

### DIFF
--- a/mkl_fft/_pydfti.pyx
+++ b/mkl_fft/_pydfti.pyx
@@ -878,7 +878,9 @@ def _remove_axis(s, axes, axis_to_remove):
     a2r = lens + axis_to_remove if axis_to_remove < 0 else axis_to_remove
 
     ss = s[:a2r] + s[a2r+1:]
-    aa = axes_normalized[:a2r] + tuple(ai - 1 for ai in axes_normalized[a2r+1:])
+    pivot = axes_normalized[a2r]
+    aa = tuple(ai if ai < pivot else ai - 1 for ai in axes_normalized[:a2r]) + \
+         tuple(ai if ai < pivot else ai - 1 for ai in axes_normalized[a2r+1:])
     return ss, aa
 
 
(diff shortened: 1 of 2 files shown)


charris:
@oleksandr-pavlyk Maybe pass that test back to NumPy, I don't think it is MKL specific.

oleksandr-pavlyk:
@charris Sure thing. I will open a PR.

PR #44, opened 2019-10-23
Feature/tls dfti cache

oleksandr-pavlyk:
This PR replaces use of global variable used to cache dfti descriptor (and required use of `threading.Lock()` to protect it from corruption in multi-threaded use) with use of thread local storage.

Consequentially, this removes the need to use `atexit` to register cache destructor.  Use of `DftiFreeDescriptor` in the function registered with `atexit` may cause a crash (like it did on Windows), since the library can not be guaranteed to still be loaded.


oleksandr-pavlyk:
@anton-malakhov @AndresGuzman-Ballen

aguzmanballen:
Wow, that's a lot of changes. Well I don't see anything nasty in there, and the CI passed. I imagine you also did some testing so assuming that you didn't encounter any issues, this should be good to merge. :+1:

PR #45, opened 2019-11-25
Add scipy1.4 backend

### LINKED ISSUE #42: Provide an interface for scipy.fft
The upcomping SciPy 1.4.0 release will include a new subpackage [`scipy.fft`] that supercedes `scipy.fftpack`. This new interface matches the behaviour of `numpy.fft` closely including real to complex `rfft`s but is different in a few ways (e.g. dtype casting).

This interface also includes a new backend mechanism (scipy/scipy#10383) that would allow `mkl_fft` to implement the `scipy.fft` interface without any monkey patching. e.g. from pyFFTW/pyFFTW#269:

```python
import scipy.fft
scipy.fft.fft([1])  # Calls scipy's own implementation

from pyfftw.interfaces import scipy_fft
scipy.fft.set_global_backend(scipy_fft)
scipy.fft.fft([1])  # Calls into pyfftw
```

See [`scipy_fft.py`] for what needs to be done on the `mkl_fft` side to support this.

It's also unfortunate that the name `mkl_fft._scipy_fft` is already taken. Any ideas on how to move forward with this?

[`scipy.fft`]: https://scipy.github.io/devdocs/fft.html
[`scipy_fft.py`]: https://github.com/pyFFTW/pyFFTW/blob/master/pyfftw/interfaces/scipy_fft.py

oleksandr-pavlyk:
Closes #42 

@peterbell10

### DIFF
--- a/mkl_fft/_scipy_fft_backend.py
+++ b/mkl_fft/_scipy_fft_backend.py
@@ -1,0 +1,133 @@
+#!/usr/bin/env python
+# Copyright (c) 2019, Intel Corporation
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+#     * Redistributions of source code must retain the above copyright notice,
+#       this list of conditions and the following disclaimer.
+#     * Redistributions in binary form must reproduce the above copyright
+#       notice, this list of conditions and the following disclaimer in the
+#       documentation and/or other materials provided with the distribution.
+#     * Neither the name of Intel Corporation nor the names of its contributors
+#       may be used to endorse or promote products derived from this software
+#       without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+# AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+# DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE
+# FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+# DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+# SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+# OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+from . import _pydfti
+import mkl
+
+import scipy.fft as _fft
+
+# Complete the namespace (these are not actually used in this module)
+from scipy.fft import (
+    dct, idct, dst, idst, dctn, idctn, dstn, idstn,
+    hfft2, ihfft2, hfftn, ihfftn,
+    fftshift, ifftshift, fftfreq, rfftfreq,
+    get_workers, set_workers
+)
+
+__all__ = ['fft', 'ifft', 'fft2', 'ifft2', 'fftn', 'ifftn',
+           'rfft', 'irfft', 'rfft2', 'irfft2', 'rfftn', 'irfftn',
+           'hfft', 'ihfft', 'hfft2', 'ihfft2', 'hfftn', 'ihfftn',
+           'dct', 'idct', 'dst', 'idst', 'dctn', 'idctn', 'dstn', 'idstn',
+           'fftshift', 'ifftshift', 'fftfreq', 'rfftfreq', 'get_workers',
+           'set_workers', 'next_fast_len']
+
+__ua_domain__ = 'numpy.scipy.fft'
+__implemented = dict()
+
+def __ua_function__(method, args, kwargs):
+    """Fetch registered UA function."""
+    fn = _implemented.get(method, None)
+    if fn is None:
+        return NotImplemented
+    return fn(*args, **kwargs)
+
+def _implements(scipy_func):
+    """Decorator adds function to the dictionary of implemented UA functions"""
+    def inner(func):
+        _implemented[scipy_func] = func
+        return func
+
+    return inner
+
+
+def _unitary(norm):
+    if norm not in (None, "ortho"):
+        raise ValueError("Invalid norm value %s, should be None or \"ortho\"."
+                         % norm)
+    return norm is not None
+
+
+@_implements(_fft.fft)
+def fft(x, n=None, axis=-1, norm=None, overwrite_x=False, workers=None):
+    output = _pydfti.fft(x, n=n, axis=axis, overwrite_x=overwrite_x)
+    if _unitary(norm):
+        output *= 1 / sqrt(output.shape[axis])
+    return output
+
+
+@_implements(_fft.ifft)
+def ifft(x, n=None, axis=-1, norm=None, overwrite_x=False, workers=None):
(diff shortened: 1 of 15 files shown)


oleksandr-pavlyk:
```
$ MKL_VERBOSE=1 python
Python 3.7.5 (default, Nov 23 2019, 04:02:01)
[GCC 7.3.0] :: Intel(R) Corporation on linux
Type "help", "copyright", "credits" or "license" for more information.
Intel(R) Distribution for Python is brought to you by Intel Corporation.
Please check out: https://software.intel.com/en-us/python-distibution
>>> import scipy.fft
mkl-service + Intel(R) MKL: THREADING LAYER: (null)
mkl-service + Intel(R) MKL: setting Intel(R) MKL to use INTEL OpenMP runtime
mkl-service + Intel(R) MKL: preloading libiomp5.so runtime
MKL_VERBOSE Intel(R) MKL 2020.0 Product build 20191102 for Intel(R) 64 architecture Intel(R) Advanced Vector Extensions 2 (Intel(R) AVX2) enabled processors, Lnx 2.40GHz lp64
 intel_thread
MKL_VERBOSE SDOT(2,0x562d021b9040,1,0x562d021b9040,1) 1.84ms CNR:OFF Dyn:1 FastMM:1 TID:0  NThr:16
>>> import mkl_fft._scipy_fft_backend as mkl_be
>>> with scipy.fft.set_backend(mkl_be, only=True):
...     scipy.fft.fft([1,2,3,4])
...
MKL_VERBOSE FFT(dcfi4,bScale:0.25,tLim:1,desc:0x562d022cf680) 7.77us CNR:OFF Dyn:1 FastMM:1 TID:0  NThr:16
array([10.+0.j, -2.+2.j, -2.+0.j, -2.-2.j])
>>> scipy.fft.fft([1,2,3,4])
array([10.-0.j, -2.+2.j, -2.-0.j, -2.-2.j])
>>> quit()
```

mkl_fft/_scipy_fft_backend.py
```diff
@@ -0,0 +1,133 @@
...
+# Complete the namespace (these are not actually used in this module)
+from scipy.fft import (
+    dct, idct, dst, idst, dctn, idctn, dstn, idstn,
+    hfft2, ihfft2, hfftn, ihfftn,
+    fftshift, ifftshift, fftfreq, rfftfreq,
+    get_workers, set_workers
+)
+
+__all__ = ['fft', 'ifft', 'fft2', 'ifft2', 'fftn', 'ifftn',
+           'rfft', 'irfft', 'rfft2', 'irfft2', 'rfftn', 'irfftn',
```
peterbell10: I don't see `rfft` implemented anywhere? It should match the numpy output format
https://github.com/IntelPython/mkl_fft/blob/aeb56b700353365904d6b8cf1c9cb0e2c5a9e90a/mkl_fft/_numpy_fft.py#L335

mkl_fft/_scipy_fft_backend.py
```diff
@@ -0,0 +1,133 @@
...
+
+def _unitary(norm):
+    if norm not in (None, "ortho"):
+        raise ValueError("Invalid norm value %s, should be None or \"ortho\"."
+                         % norm)
+    return norm is not None
+
+
+@_implements(_fft.fft)
+def fft(x, n=None, axis=-1, norm=None, overwrite_x=False, workers=None):
```
peterbell10: It's a shame that you ignore the workers argument everywhere. Could something be done with `mkl_set_num_theads`?

mkl_fft/_scipy_fft_backend.py
```diff
@@ -0,0 +1,133 @@
...
+    output = _pydfti.fft(x, n=n, axis=axis, overwrite_x=overwrite_x)
```
peterbell10: `scipy.fft` upcasts `np.float16` in the same way as `scipy.fftpack`. Might need the upcast function?
https://github.com/IntelPython/mkl_fft/blob/aeb56b700353365904d6b8cf1c9cb0e2c5a9e90a/mkl_fft/_scipy_fft.py#L34

mkl_fft/_scipy_fft_backend.py
oleksandr-pavlyk: It need not be the case. I can support `workers` by means of using `mkl-service`, using `domain_set_num_threads` ([ref](https://software.intel.com/en-us/mkl-developer-reference-c-mkl-domain-set-num-threads)). 

The only thing I am unsure of, is how to make certain setting and restoring number of threads per domain is interrupt free (so the a user inadvertently ends up with MKL in state where the number of threads for FFT domain has been reset from the previous value). 

Will the following guarantee that the number of threads be reset even in the case of keyboard interruption?

```
try:
    saved = mkl.domain_get_max_threads(domain='fft');
    mkl.domain_set_num_threads(workers, domain='fft')
    res = mkl_fft.fft(inputs..)
finally:
    mkl.domain_set_num_threads(saved, domain='fft')
return res
```

mkl_fft/_scipy_fft_backend.py
oleksandr-pavlyk: Thank you, I will fill this omission.

mkl_fft/_scipy_fft_backend.py
peterbell10: I think it should be okay, python generally raises a `KeyboardInterrupt` exception so the `finally` clause should be run.

This SO post has some cautionary remarks though:
https://stackoverflow.com/questions/41280318/will-python-execute-finally-block-after-receiving-ctrlc

oleksandr-pavlyk: @peterbell10 Thank you for your feedback. I think I implemented all of it now.

mkl_fft/_scipy_fft_backend.py
oleksandr-pavlyk: Thank you. I added the call.

mkl_fft/_scipy_fft_backend.py
```diff
@@ -0,0 +1,265 @@
...
+    s = x.shape
+    if axes is None:
+        return x.size
+    return prod([s[ai] for ai in axes])
+
+
+def _workers_to_num_threads(w):
+    if w is None:
+        return mkl.domain_get_max_threads(domain='fft')
+    return int(w)
```
peterbell10: `workers` has a slightly more complicated meaning than literally just the number of threads. It can also be negative e.g. -1 for all threads, -2 for all but 1. Ideally, `mkl_fft` should respect the `fft.set_workers` context manager which controls the default number of threads. You can use `fft.get_workers()` to access the current value. 

`pyfftw` implements all this with a `_workers_to_threads` helper function:
https://github.com/pyFFTW/pyFFTW/blob/011a808c4e480f5674d79113bfb10662c103b3d9/pyfftw/interfaces/scipy_fft.py#L100-L115

oleksandr-pavlyk: Thank you @peterbell10. 

Unfortunately Intel(R) MKL's assessment of the maximal number of threads it can run disagrees with that of `os.cpu_count()`:

```
(t_scipy-1.4.0) [11:46:11 linbox mkl_fft]$ ipython
Python 3.7.5 (default, Nov 23 2019, 04:02:01)
Type 'copyright', 'credits' or 'license' for more information
IPython 7.11.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import os, mkl

In [2]: ( os.cpu_count(), mkl.get_max_threads() )
Out[2]: (32, 16)
```

And MKL would clip large value to the maximum number of threads possible:

```
In [3]: mkl.set_num_threads(30)
Out[3]: 16
```

Moreover, the `mkl.get_max_threads()` can be controlled via environmental variable:

```
(t_scipy-1.4.0) [11:49:08 linbox mkl_fft]$ ipython -c "import mkl; print(mkl.get_max_threads())"
16
(t_scipy-1.4.0) [11:49:12 linbox mkl_fft]$ MKL_NUM_THREADS=8 ipython -c "import mkl; print(mkl.get_max_threads())"
8
```

So I will be using `mkl.get_max_threads()` to obtain the maximum value relative to which to interpret the negative `workers` value.

oleksandr-pavlyk: Added in #47

peterbell10: >I will be using mkl.get_max_threads() to obtain the maximum value relative to which to interpret the negative workers value.

That sounds a bit problematic to me since a library using the `scipy.fft` interface can't really be sure if a negative `workers` value is valid or not. Perhaps it would be better to follow `os.cpu_count()`, let `mkl.set_num_threads` clip the value and just give a warning?

cc @larsoner, @rgommers any thoughts on this?

oleksandr-pavlyk: I made that change in #47

PR #54, opened 2020-12-19
Fix global side effects of _scipy_fft_backend

### LINKED ISSUE #53: _scipy_fft_backend has global side-effects
Here is a fairly minimal reproducer:
```python
In [1]: import mkl, numpy as np
   ...: import mkl_fft
   ...: from mkl_fft._scipy_fft_backend import fft as scipy_fft
   ...:
   ...: x = np.random.rand(100, 100, 100).astype(np.cdouble)
   ...: mkl.set_num_threads(1)
   ...: %timeit scipy_fft(x, workers=8)
   ...: 
   ...: mkl.set_num_threads(8)
   ...: %timeit mkl_fft.fft(x)
320 µs ± 420 ns per loop (mean ± std. dev. of 7 runs, 1000 loops each)
3.36 ms ± 1.38 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

I would expect the `mkl_fft.fft` call to use 8 threads and so be as fast as `scipy_fft` with `workers=8`. What actually happens is that `scipy_fft` set the fft domain thread count to 1 and the domain has higher precedence than the global thread setting.

https://github.com/IntelPython/mkl_fft/blob/4d8cc2a07778ae8bac04d0af3c1079921b2115b2/mkl_fft/_scipy_fft_backend.py#L162-L163

peterbell10:
Should fix #53. I can't test though because `mkl_fft` won't build locally for me.

### DIFF
--- a/mkl_fft/_scipy_fft_backend.py
+++ b/mkl_fft/_scipy_fft_backend.py
@@ -153,14 +153,13 @@ class Workers:
 
     def __enter__(self):
         try:
-            mkl.domain_set_num_threads(self.n_threads, domain='fft')
+            self.prev_num_threads = mkl.set_num_threads_local(self.n_threads)
         except:
             raise ValueError("Class argument {} result in invalid number of threads {}".format(self.workers, self.n_threads))
 
     def __exit__(self, *args):
-        # restore default
-        n_threads = _hardware_counts.get_max_threads_count()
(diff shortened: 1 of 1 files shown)


oleksandr-pavlyk (approved):
LGTM! Thank you

PR #56, opened 2021-02-24
Fixed issue #48

### LINKED ISSUE #48: fftn on multiple arrays faster using python loop
Running `mkl_fft.fftn` on an array with shape (70, 196, 150, 24) only on the last 3 axes is more than 2 times slower than running the transform on each of the 70 sub-arrays individually using a python loop.

This is unexpected as one would assume the loop should actually run faster inside of mkl.

Simple example:
```
import mkl_fft
import time

shape = (70, 196, 150, 24)
array = np.random.random(shape) + 1j * np.random.random(shape)

def transform(array):
    result = np.empty(array.shape, dtype=complex)
    for ii, arr in enumerate(array):
        result[ii] = mkl_fft.fftn(arr)
    return result

t0 = time.time()
a = mkl_fft.fftn(array, axes=(1, 2, 3))
t1 = time.time()
b = transform(array)
t2 = time.time()

print('fftn on full array: {:.0f} ms'.format(1000*(t1 - t0)))
print('loop of fftn on subarray: {:.0f} ms'.format(1000*(t2 - t1)))
print(np.allclose(a, b))
```
On my machine this returns:
```
fftn on full array: 1359 ms
loop of fftn on subarray: 619 ms
True
```
I also verified the timings with `timeit` instead of `time.time`.

oleksandr-pavlyk:
Closes #48 

```
(bld_npy_120) [10:26:23 vmlin03 mkl_fft]$ python issue48.py
fftn on full array: 680 ms
loop of fftn on subarray: 630 ms
True
(bld_npy_120) [10:26:32 vmlin03 mkl_fft]$ python issue48.py
fftn on full array: 656 ms
loop of fftn on subarray: 677 ms
True
```

Also fixes typo in  `_numpy_fft.py` introduced in the previous commit.

@AndresGuzman-Ballen @mvandeput

### DIFF
--- a/mkl_fft/_numpy_fft.py
+++ b/mkl_fft/_numpy_fft.py
@@ -1,4 +1,4 @@
-g#!/usr/bin/env python
+#!/usr/bin/env python
 # Copyright (c) 2017-2019, Intel Corporation
 #
 # Redistribution and use in source and binary forms, with or without
--- a/mkl_fft/_pydfti.pyx
+++ b/mkl_fft/_pydfti.pyx
@@ -912,6 +912,48 @@ def _iter_fftnd(a, s=None, axes=None, function=fft, overwrite_arg=False, scale_f
     return a
 
 
+def flat_to_multi(ind, shape):
+    nd = len(shape)
+    m_ind = [-1] * nd
+    j = ind
+    for i in range(nd):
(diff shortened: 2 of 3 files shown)


AndresGuzman-Ballen:
👍

PR #78, opened 2023-03-13
Raise NotImplementedError exception instead of ValueError for non-default plan parameters in scipy fft backend

DenisScherbakov:
Raise `NotImplementedError` instead of `ValueError` when `plan` parameter passed to `mkl_fft.interfaces.scipy_fft` functions has a non-default value.


oleksandr-pavlyk (approved):

oleksandr-pavlyk:
You also need to bump the version in `conda-recipe/meta.yaml` please

PR #84, opened 2023-07-18
Support Cython 3.0

ekimd:
_capsule_destructor can no longer be implicitly converted to PyCapsule_Destructor unless it is marked as not throwing an exception.


ekimd:
While I'm not 100% certain, based on some of my reading I believe that the exception would be swallowed in <=0.29.*.  
In general, destructors shouldn't throw exceptions otherwise unpredictable behavior can result.  Line 57 should perhaps be changed to a simple print instead, or something to that effect.

oleksandr-pavlyk (approved):

PR #100, opened 2024-05-25
Changes to support NumPy 2.0

oleksandr-pavlyk:
Changes to support NumPy 2.0

Use np.lib.NumpyVersion to account for change in namespace of some functions. Replaced import from numpy.core with import from numpy main namespace, where appropriate.

Replace np.longcomplex with np.clongdouble.

Fixed import of NumPy C-API in _pydfti extension. This follows https://numpy.org/devdocs/reference/c-api/array.html#including-and-importing-the-c-api

Separate translation units, like mklfft.c, need to include arrayobject.h after defining NO_IMPORT_ARRAY preprocessor variable. The extension must be compiled with PY_ARRAY_UNIQUE_SYMBOL set to artibrary value, but same for all translation units.

### DIFF
--- a/.github/workflows/build_pip.yaml
+++ b/.github/workflows/build_pip.yaml
@@ -1,4 +1,4 @@
-name: Editable build using pip
+name: Editable build using pip and pre-release NumPy
 
 on: push
 
--- a/.github/workflows/build_pip.yaml
+++ b/.github/workflows/build_pip.yaml
@@ -34,22 +34,21 @@ jobs:
       - uses: conda-incubator/setup-miniconda@v3
         with:
           use-mamba: true
-          mamba-version: "*"
+          miniforge-variant: Mambaforge
+          miniforge-version: latest
           channels: conda-forge
           activate-environment: test
           python-version: ${{ matrix.python }}
 
       - name: Install MKL
         run: |
-          conda install -c conda-forge mkl-devel mkl-service --override-channels
-          conda activate test
+          conda install mkl-devel mkl-service
           python -c "import sys; print(sys.executable)"
           which python
           python -c "import mkl; print(mkl.__file__)"
 
       - name: Build conda package
         run: |
-          conda activate test
           pip install --no-cache-dir cython pytest hypothesis
           pip install --no-cache-dir numpy ${{ matrix.use_pre }}
           echo "CONDA_PREFFIX is '${CONDA_PREFIX}'"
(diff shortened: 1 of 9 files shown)


.github/workflows/build_pip.yaml
```diff
@@ -0,0 +1,58 @@
...
+        run: |
+          conda install -c conda-forge mkl-devel mkl-service --override-channels
+          conda activate test
+          python -c "import sys; print(sys.executable)"
+          which python
+          python -c "import mkl; print(mkl.__file__)"
+
+      - name: Build conda package
+        run: |
+          conda activate test
```
antonwolfy, line 52: I believe we no need to activate the env when `conda-incubator/setup-miniconda` is used

.github/workflows/build_pip.yaml
```diff
@@ -0,0 +1,58 @@
...
+          use-mamba: true
+          mamba-version: "*"
+          channels: conda-forge
+          activate-environment: test
+          python-version: ${{ matrix.python }}
+
+      - name: Install MKL
+        run: |
+          conda install -c conda-forge mkl-devel mkl-service --override-channels
+          conda activate test
```
antonwolfy, line 45: Are we going to use the "mamba" command here, since `use-mamba: true` was specified?
```suggestion
          mamba install mkl-devel mkl-service
```
I believe we don't need to activate the env and to pass the channels, because that was defined by default configuration in `conda-incubator/setup-miniconda` step.

.github/workflows/build_pip.yaml
```diff
@@ -0,0 +1,58 @@
...
+          sudo apt-get install jq
+
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - uses: conda-incubator/setup-miniconda@v3
+        with:
+          use-mamba: true
+          mamba-version: "*"
```
antonwolfy, line 37: Would it be helpful to switch to `Mambaforge` if we are going to use mamba?
```suggestion
          miniforge-variant: Mambaforge
          miniforge-version: latest
          use-mamba: true
```

oleksandr-pavlyk:
Thanks @antonwolfy , I have applied suggested hanges

antonwolfy (approved):
I have no more comments. LGTM!

ndgrigorian (approved):
LGTM!

Issue #109, opened 2024-07-18
Incorrect result for integer input


vtavana:
Following example returns correct result for float dtype (results similar to stock NumPy) but incorrect result for integer dtype:
```
#  Name        Version   Build              Channel
# ────────────────────────────────────────────────────
#  numpy       1.26.4   py311h689b997_3  intel  
#  numpy-base  1.26.4   py311h913173e_3  intel
#  mkl_fft  1.3.8    py311h977b55c_72  intel

import numpy
b=numpy.array([[5, 7, 6, 5], [4, 6, 4, 8], [9, 3, 7, 5]], dtype=numpy.float32)
res1 = numpy.fft.fftn(b, axes=(0,), s=None, norm="forward")
res1
# array([[ 6.        +0.j       ,  5.3333335 +0.j       , # correct result
#         5.666667  +0.j       ,  6.        +0.j       ],
#       [-0.5       +1.4433756j,  0.8333334 -0.8660254j,
#         0.16666667+0.8660254j, -0.5       -0.8660254j],
#       [-0.5       -1.4433756j,  0.8333334 +0.8660254j,
#         0.16666667-0.8660254j, -0.5       +0.8660254j]], dtype=complex64)

import numpy
b=numpy.array([[5, 7, 6, 5], [4, 6, 4, 8], [9, 3, 7, 5]])
res2 = numpy.fft.fftn(b, axes=(0,), s=None, norm="forward")
res2
# array([[10.39230485+0.j ,  9.23760431+0.j ,  9.81495458+0.j , # incorrect result
#        10.39230485+0.j ],
#       [-0.8660254 +2.5j,  1.44337567-1.5j,  0.28867513+1.5j,
#       -0.8660254 -1.5j],
#       [-0.8660254 -2.5j,  1.44337567+1.5j,  0.28867513-1.5j,
#        -0.8660254 +1.5j]])
```
Note that the result of integer `res2` is 1.73 times of `res1`. So, this seems to be an issue in scaling?
For `norm="ortho"`, the result are different as well but for `norm=None` or `"backward"`, the result are the same.

PR #111, opened 2024-07-18
Fix gh-109

oleksandr-pavlyk:
Resolves gh-109

Fix oversight in fallback of `fftn` to iterations of `fft1d` in scale computation. Added test to verify the fix

### DIFF
--- a/mkl_fft/tests/test_fftnd.py
+++ b/mkl_fft/tests/test_fftnd.py
@@ -223,8 +223,8 @@ def test_gh109():
     b_int = np.array([[5, 7, 6, 5], [4, 6, 4, 8], [9, 3, 7, 5]], dtype=np.int64)
     b = np.asarray(b_int, dtype=np.float32)
 
-    r1 = mkl_fft.fftn(b, shape=None, axes=(0,), overwrite_x=False)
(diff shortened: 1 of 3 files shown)


mkl_fft/tests/test_fftnd.py
```diff
@@ -217,3 +217,14 @@ def test_scale_nd_axes(self):
...
         r_tol, a_tol = _get_rtol_atol(X)
         assert_allclose(f, 5*f_scale, rtol=r_tol, atol=a_tol)
+
+
+def test_gh109():
+    b_int = np.array([[5, 7, 6, 5], [4, 6, 4, 8], [9, 3, 7, 5]], dtype=np.int64)
+    b = np.asarray(b_int, dtype=np.float32)
+
+    r1 = mkl_fft.fftn(b, shape=None, axes=(0,), overwrite_x=False)
+    r2 = mkl_fft.fftn(b_int, shape=None, axes=(0,), overwrite_x=False)
```
vtavana, line 227: I have checked and the issue is resolve. However, it is better to update the test with `forward_scale` other than default value 1.0. With the default value, both cases returns the same result.

oleksandr-pavlyk: Yes, my mistake. I pushed the change. Thanks for catching this @vtavana

vtavana (approved):

Issue #107, opened 2024-07-03
HTTP 403 FORBIDDEN for channel intel <https://conda.anaconda.org/intel>


rootsmusic:
When I try to run `conda install -c intel mkl_fft`, it shows the error:
> Channels:
>  - intel
>  - conda-forge
>  - defaults
>
> Platform: win-64
> Collecting package metadata (repodata.json): failed
> 
> UnavailableInvalidChannel: HTTP 403 FORBIDDEN for channel intel <https://conda.anaconda.org/intel>
> 
> The channel is not accessible or is invalid.

skapin:
yes, its down. i hope its note definitive

https://community.intel.com/t5/Intel-Integrated-Performance/Problems-installing-with-conda-HTTP-403-FORBIDDEN/td-p/1611876

https://community.intel.com/t5/oneAPI-Registration-Download/Disappearance-of-intel-conda-channel/td-p/1614250

skapin:
Use this  


https://www.intel.com/content/www/us/en/docs/oneapi/installation-guide-linux/2024-2/conda.html#GUID-BAA00723-86B2-4E54-9343-E817171F8A39

rootsmusic:
Thanks @skapin, I couldn't find Intel's feedstock on Github to open this issue.

skapin:
But there is not all package.... like intel::numpy... for me, it's fucked up all my customer conda build and workflow.... this is so garbage ....

oleksandr-pavlyk:
Since the README has been updated to use "-c https://software.repos.intel.com/python/conda" instead of "-c intel", I think this issue can be closed now.

PR #112, opened 2024-07-19
Replace intel channel

oleksandr-pavlyk:
This PR updates README.md to update installation instruction using "https://software.repos.intel.com/python/conda" as channel to replace disable "-c intel" as per experience and workaround reported in gh-107

Workflow script has also been updated accordingly.


ekomarova (approved):
Looks good, thank you

PR #114, opened 2024-07-24
Bumped up tolerance for test_numpy_fft, test_scipy_fft

oleksandr-pavlyk:
Bumped up tolerance for `test_numpy_fft`, `test_scipy_fft`

These two tests were failing for double precision floating point numbers with MKL 2024.2


vtavana (approved):
Thank you, @oleksandr-pavlyk!

PR #118, opened 2024-08-07
Add conda recipe cf and workflow

oleksandr-pavlyk:
Changes to add workflow to build with NumPy 2.0 using conda-forge channel only, and test with NumPy prior and post 2.0 release.


vtavana (approved):
Thank you, @oleksandr-pavlyk!

PR #123, opened 2024-09-05
Remove the wheels installation with anaconda from the instruction and add Intel channel for installing wheels

ekomarova:


oleksandr-pavlyk (approved):
Looks good to me

PR #130, opened 2025-03-06
update wokflow

vtavana:
In this PR, the workflow is updated 1) to account for conda-build (after 25.1.0) moving to producing .conda files instead of tarballs, and 2) to use `conda-remove-defaults` option which enables postprocessing channels list to remove 'defaults' if it was added implicitly to the 'channels' setting.


ndgrigorian (approved):
LGTM, though, I am curious why we choose to specifically build FFT with NumPy 2 and not umath or random

antonwolfy (approved):

ekomarova (approved):

PR #131, opened 2025-03-12
migrate from `setup.py` to `pyproject.toml`

vtavana:
Migrating from `setup.py` to `pyproject.toml` to avoid deprecated warning.

>>> DEPRECATION: Legacy editable install of mkl_fft==1.3.11 from file:~/repos/mkl_fft (setup.py develop) is deprecated. pip 25.1 will enforce this behaviour change. A possible replacement is to add a pyproject.toml or enable --use-pep517, and use setuptools >= 64. If the resulting installation is not behaving as expected, try using --config-settings editable_mode=compat. Please consult the setuptools documentation for more information. Discussion can be found at https://github.com/pypa/pip/issues/11457

In addition, `required_python` version is updated to be `3.9>=,<3.13`, numpy version is updated to be `>1.26.4` and also support for macOS is removed since [OneAPI support discontinued](https://www.intel.com/content/www/us/en/docs/oneapi/installation-guide-macos/2024-0/overview.html).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -58,7 +58,7 @@ classifiers = [
     "Operating System :: Unix",
     "Operating System :: MacOS"
 ]
-dependencies = ["numpy >=1.16", "mkl", "mkl-service"]
+dependencies = ["numpy >=1.26.4", "mkl", "mkl-service"]
 
 [project.optional-dependencies]
 test = ["pytest"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -56,7 +56,6 @@ classifiers = [
     "Operating System :: Microsoft :: Windows",
     "Operating System :: POSIX",
     "Operating System :: Unix",
-    "Operating System :: MacOS"
 ]
 dependencies = ["numpy >=1.26.4", "mkl", "mkl-service"]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,78 @@
+# Copyright (c) 2025, Intel Corporation
+#
+# Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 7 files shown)


pyproject.toml
```diff
@@ -0,0 +1,78 @@
...
+[build-system]
+requires = ["setuptools>=64", "Cython", "numpy"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "mkl_fft"
+dynamic = ["version"]
+description = "MKL-based FFT transforms for NumPy arrays"
+readme = { file = "README.md", content-type = "text/markdown" }
+requires-python = ">=3.7"
```
antonwolfy, line 35: Should it be bumped? Or do we support pythons 3,7 and 3,8 still?
```suggestion
requires-python = ">=3.9"
```

pyproject.toml
```diff
@@ -0,0 +1,78 @@
...
+    "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: Implementation :: CPython",
+    "Topic :: Software Development",
+    "Topic :: Scientific/Engineering",
+    "Operating System :: Microsoft :: Windows",
+    "Operating System :: POSIX",
+    "Operating System :: Unix",
+    "Operating System :: MacOS"
+]
+dependencies = ["numpy >=1.16", "mkl", "mkl-service"]
```
ekomarova, line 61: Is this restriction `numpy >=1.16` correct? We are testing essentially only 1.26.4 and can be sure that it works for >=1.26.4, but I'm not sure about the lower versions

ekomarova (approved):

PR #139, opened 2025-03-20
 raise warning similar to numpy >= 2.0

vtavana:
In this PR, numpy interface is updated to raise similar warning to numpy >= 2.0.

### DIFF
--- a/mkl_fft/_numpy_fft.py
+++ b/mkl_fft/_numpy_fft.py
@@ -1249,11 +1249,9 @@ def rfftn(a, s=None, axes=None, norm=None):
         fsc = 1.0
     elif norm == "forward":
         x = asanyarray(x)
-        s, axes = _cook_nd_args(x, s, axes)
         fsc = frwd_sc_nd(s, x.shape)
     else:
         x = asanyarray(x)
-        s, axes = _cook_nd_args(x, s, axes)
         fsc = sqrt(frwd_sc_nd(s, x.shape))
 
     return trycall(
--- a/mkl_fft/_numpy_fft.py
+++ b/mkl_fft/_numpy_fft.py
@@ -73,6 +73,7 @@ __all__ = [
 import re
 import warnings
 
+import numpy as np
 from numpy import array, asanyarray, conjugate, prod, sqrt, take
 
 from . import _float_utils
@@ -701,7 +702,7 @@ def _cook_nd_args(a, s=None, axes=None, invreal=False):
         shapeless = False
     s = list(s)
     if axes is None:
-        if not shapeless:
+        if not shapeless and np.__version__ >= "2.0":
             msg = (
                 "`axes` should not be `None` if `s` is not `None` "
                 "(Deprecated in NumPy 2.0). In a future version of NumPy, "
@@ -716,7 +717,7 @@ def _cook_nd_args(a, s=None, axes=None, invreal=False):
         raise ValueError("Shape and axes have different lengths.")
     if invreal and shapeless:
         s[-1] = (a.shape[axes[-1]] - 1) * 2
-    if None in s:
+    if None in s and np.__version__ >= "2.0":
         msg = (
             "Passing an array containing `None` values to `s` is "
             "deprecated in NumPy 2.0 and will raise an error in "
--- a/mkl_fft/_numpy_fft.py
+++ b/mkl_fft/_numpy_fft.py
@@ -74,35 +74,44 @@ import re
 import warnings
 
 import numpy as np
-from numpy import array, asanyarray, conjugate, prod, sqrt, take
+from numpy import array, conjugate, prod, sqrt, take
 
 from . import _float_utils
 from . import _pydfti as mkl_fft  # pylint: disable=no-name-in-module
 
 
+def _compute_fwd_scale(norm, n, shape):
+    _check_norm(norm)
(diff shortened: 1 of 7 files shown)


mkl_fft/_numpy_fft.py
```diff
@@ -1215,16 +1243,15 @@ def rfftn(a, s=None, axes=None, norm=None):
     """
     _check_norm(norm)
     x = _float_utils.__downcast_float128_array(a)
+    s, axes = _cook_nd_args(x, s, axes)
```
ndgrigorian, line 1246: was the `x = asanyarray(x)` call needed before this though?

mkl_fft/_numpy_fft.py
```diff
@@ -1369,16 +1396,15 @@ def irfftn(a, s=None, axes=None, norm=None):
     """
     _check_norm(norm)
     x = _float_utils.__downcast_float128_array(a)
+    s, axes = _cook_nd_args(x, s, axes, invreal=True)
```
ndgrigorian, line 1399: same as the other comment

mkl_fft/_pydfti.pyx
```diff
@@ -1200,7 +1200,7 @@ def rfftn(x, s=None, axes=None, fwd_scale=1.0):
             a = _fix_dimensions(a, tuple(ss), axes)
         if len(set(axes)) == len(axes) and len(axes) == a.ndim and len(axes) > 2:
             ss, aa = _remove_axis(s, axes, -1)
-            ind = [slice(None,None,1),] * len(s)
+            ind = [slice(None, None, 1),] * len(s)
```
ndgrigorian, line 1203: did linter miss this somehow? Just surprised it's here

mkl_fft/tests/test_pocketfft.py
```diff
@@ -519,14 +519,14 @@ def test_s_axes_none(self, op):
...
-    @pytest.mark.skip("no warning is raised in mkl_ftt")
+    @pytest.mark.skipif(np.__version__ < "2.0", reason="Requires numpy >= 2.0")
     @pytest.mark.parametrize("op", [mkl_fft.fft2, mkl_fft.ifft2])
     def test_s_axes_none_2D(self, op):
         x = np.arange(100).reshape(10, 10)
         with pytest.warns(match="`axes` should not be `None` if `s`"):
             op(x, s=(-1, 5), axes=None)
 
-    @pytest.mark.skip("no warning is raised in mkl_ftt")
+    @pytest.mark.skipif(np.__version__ < "2.0", reason="Requires numpy >= 2.0")
```
ndgrigorian, line 529: if this keeps coming up during the NumPy 2 transition, maybe a dedicated pytest marker should be made for it

ndgrigorian: doesn't have to be in this PR though

mkl_fft/_numpy_fft.py
vtavana: `x = _float_utils.__downcast_float128_array(a)` already converts input to a `numpy` array, so `x = asanyarray(x)` is not needed. Overall, updated the code to get rid of duplications.

ndgrigorian: should the `asanyarray` calls be removed then?

vtavana: yes, I have updated the code to get rid of duplications. and as part of that I have removed `asanyarray`.

mkl_fft/_pydfti.pyx
vtavana: it is `.pyx` file and needs `cython-lint`. added in https://github.com/IntelPython/mkl_fft/pull/144

ndgrigorian (approved):
LGTM!

mkl_fft/_numpy_fft.py
```diff
@@ -688,22 +663,46 @@ def ihfft(a, n=None, axis=-1, norm=None):
...
+        msg = (
+            "Passing an array containing `None` values to `s` is "
+            "deprecated in NumPy 2.0 and will raise an error in "
+            "a future version of NumPy. To use the default behaviour "
+            "of the corresponding 1-D transform, pass the value matching "
+            "the default for its `n` parameter. To use the default "
+            "behaviour for every axis, the `s` argument can be omitted."
+        )
+        warnings.warn(msg, DeprecationWarning, stacklevel=3)
+    # use the whole input array along axis `i` if `s[i] == -1 or None`
```
antonwolfy, line 703: It seems in NumPy it is only done in case `s[i] == -1`. Was that intended to consider also `None` here? Is it a kind of bug in numpy?

vtavana: while it is not a bug in NumPy, it is intentionally different here. NumPy always calculate N-d FFT by computing a series of 1-d FFTs. And so having `s[i] = None` means `n = None` for 1D FFT along axis `i` which is a valid option. Then inside the 1-d FFT, if `n is None`, it is set to `a.shape[i]` Which is the same that is used here. In OneMKL FFT, we have the option to compute N-d FFT directly (up to 7-d on cpu and 3-d on gpu), and we need to set `s[i] = a.shape[i]` beforehand.

antonwolfy: I see, it's crystal clear now. Thank you.

PR #140, opened 2025-03-24
Support mkl_fft in virtual environment out of the box

ekomarova:
Since location of "Library\bin" in the virtual environment is not on the default search path, importing of mkl_fft and numpy fails.

This change introduces "_init_helper.py" file which implements the following logic using built-in os Python module:
1. If os.add_dll_directory attribute exists, and VIRTUAL_ENV environment variable is set, and os.path.join(os.environ["VIRTUAL_ENV"], "Library", "bin") folder exists, call os.add_dll_directory with that directory

### DIFF
--- a/mkl_fft/__init__.py
+++ b/mkl_fft/__init__.py
@@ -33,3 +33,5 @@ import mkl_fft.interfaces
 
 __all__ = ['fft', 'ifft', 'fft2', 'ifft2', 'fftn', 'ifftn', 'rfftpack', 'irfftpack'
            'rfft', 'irfft', 'rfft2', 'irfft2', 'rfftn', 'irfftn','interfaces']
+
+del _init_helper
--- a/mkl_fft/__init__.py
+++ b/mkl_fft/__init__.py
@@ -24,6 +24,7 @@
 # OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
+from . import _init_helper
 from ._pydfti import (fft, ifft, fft2, ifft2, fftn, ifftn, rfftpack, irfftpack,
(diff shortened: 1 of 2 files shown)


mkl_fft/_init_helper.py
```diff
@@ -0,0 +1,30 @@
+# Copyright 2020-2024 Intel Corporation
```
antonwolfy, line 1:
```suggestion
# Copyright 2025 Intel Corporation

```

mkl_fft/_init_helper.py
```diff
@@ -0,0 +1,30 @@
...
+import sys
+
+is_venv_win32 = (
+    sys.platform == "win32"
+    and sys.base_exec_prefix != sys.exec_prefix
+    and os.path.isfile(os.path.join(sys.exec_prefix, "pyvenv.cfg"))
+)
+
+if is_venv_win32:
+    dll_dir = os.path.join(sys.exec_prefix, "Library", "bin")
```
antonwolfy, line 26: It would be good to place a comment noting why it is required.

mkl_fft/__init__.py
```diff
@@ -24,6 +24,7 @@
 # OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
+from . import _init_helper
```
antonwolfy, line 27: Do we need to call `del _init_helper` at the end of init file?

ekomarova: I added this, although it's more of a matter of style or, do we want the module to be visible to users, but it doesn't affect the behavior.

antonwolfy (approved):
Thank you @ekomarova, LGTM!

mkl_fft/__init__.py
antonwolfy: Yes, exactly. And the main intention was to keep it hidden from the user, since it is quite sensible.

PR #142, opened 2025-03-31
add pre-commit format

vtavana:
add pre-commit format


ndgrigorian (commented):
Good change! I think it would be good to add a `.git-blame-ignore-revs` file too, though

ndgrigorian (commented):
Also, a workflow for pre-commit would align with the other projects

It would be a good idea to make similar changes in umath and random, as well as the migration to `pyproject.toml`

Maybe issues should be made for these things

antonwolfy (approved):

.pre-commit-config.yaml
```diff
@@ -0,0 +1,42 @@
...
+- repo: https://github.com/macisamuele/language-formatters-pre-commit-hooks
+  rev: v2.14.0
+  hooks:
+  - id: pretty-format-toml
+    args: [--autofix]
+
+- repo: https://github.com/pocc/pre-commit-hooks
+  rev: v1.3.5
+  hooks:
+  - id: clang-format
```
antonwolfy, line 41: I guess the default configuration will not consider `mklfft.c.src` file

.pre-commit-config.yaml
```diff
@@ -0,0 +1,42 @@
...
+    additional_dependencies:
+        - flake8-docstrings==1.7.0
+        - flake8-bugbear==24.4.26
+
+- repo: https://github.com/macisamuele/language-formatters-pre-commit-hooks
+  rev: v2.14.0
+  hooks:
+  - id: pretty-format-toml
+    args: [--autofix]
+
```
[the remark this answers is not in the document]
antonwolfy: And do we need `pylint`?

.pre-commit-config.yaml
vtavana: formatting `mklfft.c.src` causes a lot of compile error. For instance, formatting changes `@domain@` to `@domain @` which is invalid.

.pre-commit-config.yaml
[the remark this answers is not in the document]
vtavana: added `shellcheck` and `pylint`. However, `pylint` is added with `"--errors-only"` otherwise it suggests a lot of changes which needs to be addressed separately, if needed.

_vendored/conv_template.py
ndgrigorian, line 1: Should we add it as an exclude to `flake8` as well and exclude it in `.pre-commit-config.yaml`?

Since it's a vendored file, it makes sense to me to roll these changes back and exclude it from style checks

vtavana: excluded it from `isort` and `flak8`

ndgrigorian (approved):
Nothing more from me, I think this looks good, thanks for your work on this @vtavana

antonwolfy (approved):
Thank you @vtavana

PR #143, opened 2025-04-01
Use `isort` and add pre-commit workflow

ndgrigorian:
This PR proposes in addition to adding pre-commit, adding isort and a workflow for pre-commit

Also fixes files which seemed to be failing black checks and adds black and isort to `pyproject.toml`


vtavana:
included in https://github.com/IntelPython/mkl_fft/pull/142

PR #145, opened 2025-04-02
update `_scipy_fft.py`

vtavana:
In this PR, `_scipy_fft.py` file is updated to get rid of duplications. 
In addition new tests from `scipy.fft` module are added to the test suite.


pyproject.toml
```diff
@@ -50,7 +50,7 @@ classifiers = [
   "Operating System :: POSIX",
   "Operating System :: Unix"
 ]
-dependencies = ["numpy >=1.26.4", "mkl", "mkl-service"]
+dependencies = ["numpy>=1.26.4", "mkl", "mkl-service"]
```
antonwolfy, line 53: It might be helpful to use [`optional-dependencies`](https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#optional-dependencies) like
```toml
[project.optional-dependencies]
test= ["pytest", "scipy"]
```

.github/workflows/build_pip.yaml
```diff
@@ -54,6 +54,7 @@ jobs:
         run: |
           pip install --no-cache-dir cython pytest hypothesis
           pip install --no-cache-dir numpy ${{ matrix.use_pre }}
+          pip install --no-cache-dir scipy
```
antonwolfy, line 57: It seems not needed for build, but only for testing.
Do we need to move that after the build command (or part of, if possible).?

mkl_fft/_fft_utils.py
```diff
@@ -0,0 +1,51 @@
...
+# FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+# DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+# SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+# OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+import numpy as np
+
+__all__ = ["check_norm", "compute_fwd_scale"]
```
ndgrigorian, line 29: Should these still be kept as internal (i.e., underscored)?

vtavana: My practice is to use an underscore if the helper function is inside the same file. If the helper function is in a separate utility file, I don't use an underscore. However, if you think it's better to start their names with an underscore, I'm fine with that too.

ndgrigorian: I only think it's better because it makes it clear to users that the function is intended for internal use.

It's unlikely someone would try and use it in some workflow, but with internal functions, we can make API changes freely without expecting users to rely on them to be stable.

vtavana: done

mkl_fft/_float_utils.py
```diff
@@ -82,7 +82,7 @@ def __downcast_float128_array(x):
 def __supported_array_or_not_implemented(x):
```
ndgrigorian, line 82: Nit but is there any good reason for these functions to be double-underscored, as well?

Generally double-underscore is meant to prevent overwriting by subclasses, so it's not really doing anything here.

It doesn't really matter though, purely a style question.

vtavana: I agree. I'm not sure why it was used either.

ndgrigorian (approved):
Tests run so LGTM!

PR #156, opened 2025-04-10
change import order to avoid circular dependency

vtavana:
change import order in `mkl_fft/__ini__.py` to avoid circular dependency that occurs in building wheel


ekomarova:
I will test this change in internal CI and let you know.

ekomarova:
It works

ekomarova (approved):
This change looks good for me and it completely fixes the issue with import

PR #159, opened 2025-04-16
Add OpenSSF score-card workflow and badge

vtavana:
Add OpenSSF score-card workflow and badge


antonwolfy (approved):
LGTM! Thank you @vtavana

PR #160, opened 2025-04-17
extend github workflows with build with clang

vtavana:


.github/workflows/build-with-clang.yml
```diff
@@ -0,0 +1,77 @@
...
+        shell: bash -l {0}
+        run: ls /opt/intel/oneapi/compiler
+
+      - name: Build mkl_fft
+        shell: bash -l {0}
+        run: |
+          source /opt/intel/oneapi/setvars.sh
+          echo $CMPLR_ROOT
+          export CC=$CMPLR_ROOT/bin/icx
+          export CXX=$CMPLR_ROOT/bin/icpx
```
ndgrigorian, line 68: nit but is setting CXX compiler necessary here? Since only C sources exist

ndgrigorian (approved):
LGTM!

Issue #176, opened 2025-05-05
`irfft` returns `float64` for an input with dtype of `float32`


vtavana:
`irfft` returns `float64` for an input with dtype of `float32`

```python
import mkl_fft, numpy
a = numpy.ones(10, dtype="f4")

mkl_fft.irfft(a).dtype
# dtype('float64')

numpy.__version__
# '2.2.3'
numpy.fft.irfft(a).dtype
# dtype('float32')
```

PR #180, opened 2025-05-09
fix an issue for real inputs of `irfft`

vtavana:
resolves https://github.com/IntelPython/mkl_fft/issues/176

### DIFF
--- a/mkl_fft/tests/test_fft1d.py
+++ b/mkl_fft/tests/test_fft1d.py
@@ -465,5 +465,4 @@ def test_irfft_dtype(dt):
     x = np.array(rnd.random((20, 20)), dtype=dt)
     result = mkl_fft.irfft(x)
     expected = np.fft.irfft(x)
-    assert result.dtype == expected.dtype
-    assert_allclose(result, expected, rtol=1e-7, atol=1e-7)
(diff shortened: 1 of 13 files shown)


mkl_fft/tests/test_fft1d.py
```diff
@@ -457,3 +457,13 @@ def test_irfft_out_strided(axis):
...
+
+
+@requires_numpy_2
+@pytest.mark.parametrize("dt", ["i4", "f4", "f8", "c8", "c16"])
+def test_irfft_dtype(dt):
+    x = np.array(rnd.random((20, 20)), dtype=dt)
+    result = mkl_fft.irfft(x)
+    expected = np.fft.irfft(x)
+    assert result.dtype == expected.dtype
+    assert_allclose(result, expected, rtol=1e-7, atol=1e-7)
```
antonwolfy, line 469: Would it make sense to pass `strict` instead, since it was added since numpy 2.0?
```suggestion
    assert_allclose(result, expected, rtol=1e-7, atol=1e-7, strict=True)
```

antonwolfy (approved):
LGTM, a small nit/question below

PR #181, opened 2025-05-09
update handling shape and axes of scipy interface

vtavana:
In this PR,

1) SciPy interface is updated to use the same function from SciPy for handling shape and axes in N-D FFTs (Currently it uses `_cook_nd_args` function which is from NumPy, in this PR, this function is replaced with `_init_nd_shape_and_axes` which is from SciPy)

2) Although SciPy has `overwrite_x` parameter for real-to-complex and complex-to-real FFTs, [it has not been used internally](https://github.com/scipy/scipy/blob/main/scipy/fft/_pocketfft/basic.py#L60). It has not been used in `mkl_fft` as well and it is not a limitation for `mkl_fft` so docstring is updated accordingly.


antonwolfy (approved):
LGTM

PR #182, opened 2025-05-13
use `PyLong_AsLong` instead of `PyInt_AsLong`

vtavana:
[`PyInt_AsLong`](https://docs.python.org/2/c-api/int.html#:~:text=long%20PyInt_AsLong(PyObject%C2%A0*io)) is deprecated and [`PyLong_AsLong`](https://docs.python.org/3/c-api/long.html#:~:text=long%20PyLong_AsLong(PyObject%20*obj)%C2%B6) should be used with python3.


ndgrigorian (approved):
LGTM! Given that it was part of Python 2 C API I'm surprised it was ever even used

PR #185, opened 2025-05-15
Revisit usage of `overwrite_x` parameter

vtavana:


mkl_fft/interfaces/_scipy_fft.py
```diff
@@ -203,6 +203,17 @@ def _init_nd_shape_and_axes(x, shape, axes, invreal=False):
...
 
+def _use_input_as_out(x, overwrite_x):
+    """Check if the input can be used as output."""
+    if overwrite_x and np.issubdtype(x.dtype, np.complexfloating):
+        # pass input as out to overwrite it
+        out = x
+    else:
+        out = None
+
+    return out
```
ndgrigorian, line 214:
```suggestion
    if overwrite_x and np.issubdtype(x.dtype, np.complexfloating):
        # pass input as out to overwrite it
        return x
    else:
        return None
```
could be simplified like this

PR #179, opened 2025-05-09
add `README.md` to interfaces

vtavana:
In this PR,

1) Project `README.md` file is updated and a new `README.md` file is added to interfaces module. resolves https://github.com/IntelPython/mkl_fft/issues/49
2) support for `fftfreq`, `fftshift`, `ifftshift`, and `rfftfreq` is added to both NumPy and SciPy interfaces by calling back on their original implementation in NumPy and SciPy. This is done for completeness as requested by https://github.com/IntelPython/mkl_fft/issues/106
3) With this PR, To set `mkl_fft` as the backend for SciPy is only possible through `mkl_fft.interfaces.scipy_fft` as explained in the examples. resolves https://github.com/IntelPython/mkl_fft/issues/170

PR #187, opened 2025-05-17
add scipy to runtime dependency

vtavana:
In https://github.com/IntelPython/mkl_fft/pull/179, we added helper functions [directly from SciPy](https://github.com/IntelPython/mkl_fft/blob/master/mkl_fft/interfaces/scipy_fft.py#L29) for expanding the functionality of `mkl_fft.interfaces.scipy_fft` module. This addition means that `scipy` is now a runtime dependency. In this PR it is added to the project as a runtime dependency.

### DIFF
--- a/.github/workflows/conda-package-cf.yml
+++ b/.github/workflows/conda-package-cf.yml
@@ -132,7 +132,7 @@ jobs:
       - name: Install mkl_fft
         run: |
           CHANNELS="-c $GITHUB_WORKSPACE/channel ${{ env.CHANNELS }}"
-          conda create -n ${{ env.TEST_ENV_NAME }} python=${{ matrix.python_ver }} ${{ matrix.numpy }} $PACKAGE_NAME pytest scipy $CHANNELS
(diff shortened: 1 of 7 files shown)


vtavana:
@ekomarova needs this PR to be merged to ensure wheel packages function correctly.
And to have a green CI for this PR, we need to first merge https://github.com/IntelPython/mkl_fft/pull/190

@ndgrigorian @antonwolfy When you have time, please review them.

ndgrigorian:
My only concern is, does this mean that Intel NumPy can no longer be installed without SciPy, if mkl_fft is being used? Is this something we should do intentionally?

@ekomarova does infra team have concerns about this?

ndgrigorian:
@vtavana 
I think that's sensible, to avoid introducing a strict dependency.

An additional alternative is that tests that use SciPy could be skipped without it in the environment, but that isn't strictly necessary I think.

vtavana:
superseded by https://github.com/IntelPython/mkl_fft/pull/195

PR #190, opened 2025-05-21
add version and build of `mkl_fft` package when running test on Linux in CI

vtavana:
When running tests on Linux in the CI with python 3.9 and 3.10, `mkl_fft` package was installed from `https://software.repos.intel.com/python/conda` and not the local channel. This PR fixes it to always install `mkl_fft` from local channel.

In addition, the only scipy available for python 3.9 and 3.10 from `https://software.repos.intel.com/python/conda` is `scipy=1.10.1` which requires `mkl>=2024.2,<2025`. Therfore, `mkl-devel` is pinned for python 3.9 and 3.10 in `meta.yaml` file.


ndgrigorian (approved):
LGTM

PR #195, opened 2025-06-03
conditional import `scipy_fft`

vtavana:
In this PR, `mkl_fft.interface` is modified to only import `scipy_fft` if `scipy` is installed.


ndgrigorian (approved):
LGTM!

PR #217, opened 2025-08-13
Add description to the recipe to meet internal validation checks

ekomarova:
Internal validation requires field `description` in the recipe to be successful

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -48,7 +48,7 @@ about:
     home: http://github.com/IntelPython/mkl_fft
     license: BSD-3-Clause
     license_file: LICENSE.txt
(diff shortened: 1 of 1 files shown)


conda-recipe/meta.yaml
```diff
@@ -49,3 +49,12 @@ about:
     license: BSD-3-Clause
     license_file: LICENSE.txt
     summary: NumPy-based implementation of Fast Fourier Transform using Intel® oneAPI Math Kernel Library (OneMKL)
```
vtavana, line 51:
```suggestion
    summary: NumPy-based implementation of Fast Fourier Transform using Intel® oneAPI Math Kernel Library (oneMKL)
```

PR #224, opened 2025-09-10
add manual patching to `mkl_fft`

ndgrigorian:
this PR proposes implementing manual patching a la [mkl_umath](https://github.com/IntelPython/mkl_umath/blob/main/mkl_umath/src/_patch.pyx) to mkl_fft

this patching is specifically designed with [NEP-36](https://numpy.org/neps/nep-0036-fair-play.html#questions-and-answers) in mind


mkl_fft/_patch_numpy.py
```diff
@@ -0,0 +1,193 @@
+#!/usr/bin/env python
```
jharlow-intel: I think we should remove this shebang, and maybe from other files too in this repo?
Could be an issue if ran on a system that uses `python3` instead of python or something old.

Not the most important thing though

antonwolfy (approved):
LGTM! The only comment from me is that it'd be great to update the changelog.

PR #226, opened 2025-09-25
Use wrappers to complete SciPy and NumPy interface namespaces

ndgrigorian:
This PR proposes using wrappers for `fftfreq`, `fftshift`, `ifftshift`, and `rfftfreq` instead of importing them from `scipy.fft` or `numpy.fft` directly

These thin wrappers can avoid infinite recursion of dependencies, i.e., when importing Intel NumPy with Intel SciPy in the environment

### DIFF
--- a/mkl_fft/interfaces/_scipy_fft.py
+++ b/mkl_fft/interfaces/_scipy_fft.py
@@ -669,7 +669,7 @@ def fftfreq(n, d=1.0, *, xp=None, device=None):
 def rfftfreq(n, d=1.0, *, xp=None, device=None):
     """
     Return the Discrete Fourier Transform sample frequencies (for usage with
-    rfft, irfft).
+    `rfft`, `irfft`).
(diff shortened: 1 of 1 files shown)


antonwolfy (approved):
The fix LGTM! But need to add the PR to the changelog also.

ndgrigorian:
> The fix LGTM! But need to add the PR to the changelog also.

done

ndgrigorian:
> Will a change like this always require scipy as a mkl_fft dependency, or can it still be optional?

It can still be optional—since the scipy interface is only exposed in `interfaces/__init__.py`  if SciPy can be imported, this doesn't change anything (`scipy_fft.py` already needed to import it before)

alternatively, the import could be made lazy, so we could do `from scipy.fft import ffftshift` in the def of `fftshift`.

ekomarova (approved):

PR #303, opened 2026-04-07
Move build system to `meson-python`

ndgrigorian:
This PR proposes moving from `setuptools` to `meson-python` as the `mkl_fft` build system

`meson-python` is already used by NumPy and allows `setup.py` to be removed (with its logic moved into the `meson.build` script)


_vendored/process_src_template.py
```diff
@@ -0,0 +1,66 @@
...
+        type=str,
+        help="An ignored input - may be useful to add a "
+        "dependency between custom targets",
+    )
+    args = parser.parse_args()
+
+    if not args.infile.endswith('.src'):
+        raise ValueError(f"Unexpected extension: {args.infile}")
+
+    outfile_abs = os.path.join(os.getcwd(), args.outfile)
```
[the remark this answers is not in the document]
ndgrigorian: vendored file, no changes will ever be made to it.

meson.build
```diff
@@ -0,0 +1,95 @@
...
+# numpy includes
+np_dir = run_command(py,
+    ['-c', 'import numpy; print(numpy.get_include())'],
+    check: true
+).stdout().strip()
+
+inc_np = include_directories(np_dir, 'mkl_fft/src', 'mkl_fft')
+
+# compiler/linker
+cc = meson.get_compiler('c')
```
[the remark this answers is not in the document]
ndgrigorian: removing causes crashes, because the get_compiler call properly sets the compiler to use what `CC` is set to. Otherwise, it can break at runtime

conda-recipe/meta.yaml
```diff
@@ -40,7 +43,7 @@ requirements:
     run:
       - python
       - python-gil     # [py>=314]
-      - mkl-service
+      - {{ pin_compatible('mkl') }}
```
antonwolfy, line 35: The log points to missing dependency on openmp:
```bash
WARNING (mkl_fft,lib/python3.14/site-packages/mkl_fft/_pydfti.cpython-314-x86_64-linux-gnu.so): Needed DSO lib/libiomp5.so found in ['https://software.repos.intel.com/python/conda/linux-64::intel-openmp==2025.3.3=intel_30']
WARNING (mkl_fft,lib/python3.14/site-packages/mkl_fft/_pydfti.cpython-314-x86_64-linux-gnu.so): .. but ['https://software.repos.intel.com/python/conda/linux-64::intel-openmp==2025.3.3=intel_30'] not in reqs/run, (i.e. it is overlinking) (likely) or a missing dependency (less likely)
```

conda-recipe-cf/meta.yaml
```diff
@@ -15,18 +15,21 @@ requirements:
...
-      - setuptools >=77
       - mkl-devel
       - cython
       - numpy
       - wheel  >=0.41.3
     run:
       - python
       - python-gil     # [py>=314]
-      - mkl-service
+      - {{ pin_compatible('mkl') }}
```
antonwolfy, line 32: The log points to missing dependency on openmp:
```bash
WARNING (mkl_fft,lib/python3.14/site-packages/mkl_fft/_pydfti.cpython-314-x86_64-linux-gnu.so): Needed DSO lib/libomp.so found in ['conda-forge/linux-64::llvm-openmp==22.1.2=h4922eb0_0']
WARNING (mkl_fft,lib/python3.14/site-packages/mkl_fft/_pydfti.cpython-314-x86_64-linux-gnu.so): .. but ['conda-forge/linux-64::llvm-openmp==22.1.2=h4922eb0_0'] not in reqs/run, (i.e. it is overlinking) (likely) or a missing dependency (less likely)
```

conda-recipe/meta.yaml
ndgrigorian: fixed this

vlad-perevezentsev:
The last commit is experimental https://github.com/IntelPython/mkl_fft/pull/303/changes/54edcb2a9e0d057e42df4cc59d09b3f444823852 that switches `conda-recipe` build scripts from `pip wheel`/`pip install` to `python -m build` aligning with `mkl-service`  to enable verbose ninja output by  `-Ccompile-args=-v`.
It also adds `python-build >=1.2.2` as a dependency

vlad-perevezentsev:
@antonwolfy @jharlow-intel 
This PR is ready for review

.github/workflows/build_pip.yml
vlad-perevezentsev: unused env variables

conda-recipe/build.sh
```diff
@@ -1,18 +1,32 @@
...
+${PYTHON} -m pip install dist/mkl_fft*.whl \
+    --no-build-isolation \
+    --no-deps \
+    --only-binary :all: \
+    --no-index \
+    --prefix "${PREFIX}" \
+    -vv
+
+# Copy wheel package
+if [[ -d "${WHEELS_OUTPUT_FOLDER}" ]]; then
```
antonwolfy, line 27: Update `.github/workflows/conda-package.yml` to upload the wheels

conda-recipe-cf/build.sh
```diff
@@ -1,6 +1,3 @@
-#!/bin/bash -x
+#!/bin/bash -ex
 
-# make sure that compiler has been sourced, if necessary
-
-export MKLROOT=${PREFIX}
-$PYTHON setup.py build --force install --old-and-unmanageable
+$PYTHON -m pip install --no-build-isolation --no-deps -Csetup-args="-Dmkl_threading=gnu_thread" .
```
antonwolfy: That forces `-Dmkl_threading=gnu_thread`, which makes `MKLConfig.cmake` link `-lgomp` (GNU libgomp). But `conda-recipe-cf/meta.yaml` ships `llvm-openmp`, which provides` libomp`/`libiomp5`, not `libgomp`. The build resolves `-lgomp` from libgcc at build time, but the declared runtime pin is the wrong OpenMP

conda-recipe-cf/bld.bat
```diff
@@ -1,3 +1,2 @@
-set MKLROOT=%PREFIX%
-%PYTHON% setup.py build --force install --old-and-unmanageable
+%PYTHON% -m pip install --no-build-isolation --no-deps .
```
antonwolfy, line 1: That passes no threading arg, so it falls back to the `meson.options` default `intel_thread` — which requires `libiomp5md` (intel-openmp), but the cf recipe ships only `llvm-openmp`. That's either a configure failure or a runtime load failure.

meson.build
```diff
@@ -0,0 +1,115 @@
...
+
+# compiler/linker
+# this looks like dead code, but it isn't: removing get_compiler call
+# can cause segfaults at runtime (most notably when building with Intel clang)
+cc = meson.get_compiler('c')
+c_args = ['-DNDEBUG']
+
+rpath_link_args = []
+if host_machine.system() != 'windows'
+    if host_machine.system() == 'darwin'
```
antonwolfy: We do not support macOS for mkl_fft

conda-recipe/build.sh
```diff
@@ -1,18 +1,32 @@
...
+#!/bin/bash -ex
+
+if [ -d "build" ]; then
+    rm -rf build
+fi
 
-export MKLROOT=$PREFIX
 export CFLAGS="-I$PREFIX/include $CFLAGS"
-export LDFLAGS="-Wl,-rpath,\$ORIGIN/../.. -Wl,-rpath,\$ORIGIN/../../.. -L${PREFIX}/lib ${LDFLAGS}"
+export LDFLAGS="-L${PREFIX}/lib ${LDFLAGS}"
```
antonwolfy: Why do we still need that?

conda-recipe-cf/meta.yaml
antonwolfy, line 11: Needs to mute conda-build warning:
```suggestion
      - blas
      - numpy
```

conda-recipe/build.sh
vlad-perevezentsev: Good catch
Done

conda-recipe-cf/meta.yaml
vlad-perevezentsev: done

conda-recipe/build.sh
vlad-perevezentsev: You are right.
It is no longer needed because meson-python resolves MKL path via `dependency('MKL', method: 'cmake')`
Removed, thank you!

meson.build
vlad-perevezentsev: Done

conda-recipe-cf/build.sh
vlad-perevezentsev: @antonwolfy I checked this in public CI
https://github.com/IntelPython/mkl_fft/actions/runs/30023599864/job/89262291762?pr=303#step:9:128 
`llvm-openmp` ships `libgomp.so.1` and `libiomp5.so` as symlinks to `libomp.so` so `gnu_thread` correctly resolves to `llvm-openmp`

conda-recipe-cf/bld.bat
vlad-perevezentsev: @antonwolfy `llvm-openmp` ships `libiomp5md.dll` as well
https://github.com/IntelPython/mkl_fft/actions/runs/30023599864/job/89262291797?pr=303#step:8:143

conda-recipe-cf/build.sh
antonwolfy: Thanks for checking in CI — confirmed, and it's an intentional, documented contract, not a coincidence.

Two small clarifications for the record:
- On Linux the `libiomp5.so` symlink ships in `llvm-openmp` itself (from LLVM's upstream `LIBOMP_INSTALL_ALIASES`), but `libgomp.so.1` is actually provided by the `_openmp_mutex=*=*_llvm` package that `llvm-openmp` force-installs via `run_exports` — the current feedstock `rm -f`'s `libgomp*` out of `llvm-openmp` itself. Net result is what you said (both resolve to `libomp.so`), just guaranteed across the `llvm-openmp` + `_openmp_mutex` pair rather than by `llvm-openmp`'s file list alone.
- This is endorsed by the [conda-forge OpenMP KB](https://conda-forge.org/docs/maintainer/knowledge_base/#openmp) ("LLVM's implementation includes GNU's symbols in addition to LLVM ones"), and `_openmp_mutex` structurally guarantees a single OpenMP runtime per env, so there's no real risk of two clashing runtimes.

The new "Check llvm-openmp linkage" step is a nice guard. Only suggestion: a one-line comment next to `-Dmkl_threading=gnu_thread` pointing at the KB, so the "why is gnu_thread safe with llvm-openmp?" reasoning isn't lost for future readers.

.github/workflows/conda-package.yml
antonwolfy: Stray `}`

conda-recipe-cf/build.sh
vlad-perevezentsev: Done, thanks

meson.build
```diff
@@ -0,0 +1,110 @@
...
+if host_machine.system() == 'linux'
+    rpath = '$ORIGIN' / '../..' + ':' + '$ORIGIN' / '../../..'
+    rpath_link_args = ['-Wl,-rpath,' + rpath]
+endif
+
+mkl_dep = dependency('MKL', method: 'cmake',
+    modules: ['MKL::MKL'],
+    cmake_args: [
+        '-DMKL_ARCH=intel64',
+        '-DMKL_LINK=dynamic',
```
antonwolfy: I wonder why we use SDL here for mkl-service and SDL also in legacy mkl_fft. Is that intentional to migrate to dynamic link?

vlad-perevezentsev: I am not sure why `sdl` was changed to `dynamic`
@ndgrigorian could you please explain this?

In any case this change is not part of the meson migration.
If needed it can be done in a separate PR
I reverted it back to sdl

conda-recipe-cf/meta.yaml
```diff
@@ -9,24 +9,29 @@ build:
...
     host:
+      - meson-python >=0.13.0
+      - cmake
+      - ninja
       - python
       - python-gil     # [py>=314]
       - pip
-      - setuptools >=77
       - mkl-devel
+      - llvm-openmp
```
antonwolfy: Do we need that dep in case of SDL link? (same in cf recept)

.github/workflows/build-with-standard-clang.yml
```diff
@@ -0,0 +1,64 @@
...
+      - name: Build mkl_fft
+        run: |
+          export CC=${{ env.COMPILER_ROOT }}/clang
+          pip install . --no-build-isolation --no-deps --verbose
+
+      - name: Run mkl_fft tests
+        run: |
+          pip install pytest mkl-service scipy
+          # mkl_fft cannot be installed in editable mode, we need
+          # to change directory before importing it and running tests
```
antonwolfy: Is that still valid?
We are using `pip install -e ...` in other workflows and that works.

.github/workflows/build-with-standard-clang.yml
```diff
@@ -0,0 +1,64 @@
...
+          cd ..
```
antonwolfy: There is actually to avoid source-tree shadowing, we need to update the above comment.

conda-recipe/build.sh
```diff
@@ -1,18 +1,31 @@
-#!/bin/bash -x
+#!/bin/bash -ex
+
+if [ -d "build" ]; then
+    rm -rf build
+fi
 
-export MKLROOT=$PREFIX
 export CFLAGS="-I$PREFIX/include $CFLAGS"
```
antonwolfy: Do we still need that?

.github/workflows/build-with-standard-clang.yml
vlad-perevezentsev: I thought editable install would be better here
Verified locally and it works so I have updated it

conda-recipe-cf/meta.yaml
vlad-perevezentsev: Removed
Thank you

conda-recipe/build.sh
vlad-perevezentsev: No. i have removed it
Thank you

.github/workflows/build-with-standard-clang.yml
vlad-perevezentsev: No longer relevant

antonwolfy (approved):
Thank you @vlad-perevezentsev 
No more comments from me, LGTM!

meson.build
ndgrigorian: Reasoning was related to the change in `mkl_umath` to specify `gnu_thread`/`intel_thread` due to the issues with large size arrays. But probably, it wasn't needed.

PR #310, opened 2026-04-14
ASV Benchmarks Integration

vchamarthi:
Adds a complete ASV benchmark suite for mkl_fft and wires it into the internal Jenkins CI pipeline.

Benchmarks added (`benchmarks/benchmarks/`):

`bench_fft1d.py` — 1-D C2C and R2C/C2R, power-of-two and non-power-of-two
`bench_fftnd.py` — 2-D and N-D C2C and R2C/C2R, square, non-square, non-power-of-two
`bench_numpy_fft.py` — full coverage of mkl_fft.interfaces.numpy_fft
`bench_scipy_fft.py` — full coverage of mkl_fft.interfaces.scipy_fft including Hermitian 2-D/N-D
`bench_memory.py` — peak RSS for 1-D, 2-D, and 3-D transforms
`__init__.py` — pins MKL_NUM_THREADS=4 on machines with ≥4 physical cores for consistent cross-machine results (A hack to keep the results consistent on random nodes until CI finds a stable benchmarking machine)
Config (`benchmarks/asv.conf.json`)

Docs (`benchmarks/README.md`): structure, coverage, threading rationale, local run commands, CI flow.


benchmarks/benchmarks/__init__.py
```diff
@@ -0,0 +1,50 @@
...
+    try:
+        with open("/proc/cpuinfo") as f:
+            content = f.read()
+        cpu_cores = int(re.search(r"cpu cores\s*:\s*(\d+)", content).group(1))
+        sockets = max(
+            len(set(re.findall(r"physical id\s*:\s*(\d+)", content))), 1
+        )
+        return cpu_cores * sockets
+    except Exception:
+        return os.cpu_count() or 1
```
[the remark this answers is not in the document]
vchamarthi: fallbacks to single threaded

benchmarks/benchmarks/bench_numpy_fft.py
```diff
@@ -0,0 +1,232 @@
...
+
+    Complex dtypes get non-zero imaginary parts for a realistic signal.
+    *shape* may be an int (1-D) or a tuple.
+    """
+    dt = np.dtype(dtype)
+    s = (shape,) if isinstance(shape, int) else shape
+    if dt.kind == "c":
+        return (rng.standard_normal(s) + 1j * rng.standard_normal(s)).astype(dt)
+    return rng.standard_normal(s).astype(dt)
+
```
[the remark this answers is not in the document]
vchamarthi: refactored

benchmarks/benchmarks/__init__.py
```diff
@@ -0,0 +1,50 @@
...
+
+def _thread_count():
+    physical = _physical_cores()
+    return str(_MIN_THREADS) if physical >= _MIN_THREADS else "1"
+
+
+_THREADS = os.environ.get("MKL_NUM_THREADS", _thread_count())
+os.environ["MKL_NUM_THREADS"] = _THREADS
+os.environ.setdefault("OMP_NUM_THREADS", _THREADS)
+os.environ.setdefault("OPENBLAS_NUM_THREADS", _THREADS)
```
ndgrigorian: should we be manipulating anything but the MKL threads? I'm not sure one way or the other, just curious what the reasoning would be or if it's expected of users

benchmarks/benchmarks/bench_fft1d.py
```diff
@@ -0,0 +1,124 @@
...
+
+_RNG_SEED = 42
+
+
+# ---------------------------------------------------------------------------
+# Complex-to-complex 1-D (power-of-two sizes)
+# ---------------------------------------------------------------------------
+
+
+class TimeFFT1D:
```
ndgrigorian:
```suggestion
class BenchFFT1D:
```

aligns with the ASV benchmarks in mkl_umath

ndgrigorian: similar to in umath, I think we should refactor the common aspects into a base class like `BenchFFT` and if we need specialized params, just subclass them

means it will be easier to maintain in the future too

benchmarks/README.md
```diff
@@ -0,0 +1,85 @@
...
+benchmarks/
+├── asv.conf.json          # ASV configuration (CI-only, no env/build settings)
+└── benchmarks/
+    ├── __init__.py        # Thread pinning (MKL_NUM_THREADS)
+    ├── bench_fft1d.py     # mkl_fft root API — 1-D transforms
+    ├── bench_fftnd.py     # mkl_fft root API — 2-D and N-D transforms
+    ├── bench_numpy_fft.py # mkl_fft.interfaces.numpy_fft — full coverage
+    ├── bench_scipy_fft.py # mkl_fft.interfaces.scipy_fft — full coverage
+    └── bench_memory.py    # Peak RSS memory benchmarks
+```
```
ndgrigorian: don't think we need this

benchmarks/README.md
```diff
@@ -0,0 +1,85 @@
...
+| File | API | Transforms |
+|------|-----|-----------|
+| `bench_fft1d.py` | `mkl_fft` | `fft`, `ifft`, `rfft`, `irfft` — power-of-two and non-power-of-two |
+| `bench_fftnd.py` | `mkl_fft` | `fft2`, `ifft2`, `rfft2`, `irfft2`, `fftn`, `ifftn`, `rfftn`, `irfftn` |
+| `bench_numpy_fft.py` | `mkl_fft.interfaces.numpy_fft` | All exported functions including Hermitian (`hfft`, `ihfft`) |
+| `bench_scipy_fft.py` | `mkl_fft.interfaces.scipy_fft` | All exported functions including Hermitian 2-D/N-D (`hfft2`, `hfftn`) |
+| `bench_memory.py` | `mkl_fft` | Peak RSS for 1-D, 2-D, and 3-D transforms |
+
+Benchmarks cover float32, float64, complex64, complex128 dtypes, power-of-two
+and non-power-of-two sizes, square and non-square/non-cubic shapes.
```
ndgrigorian: I would add this to the table instead, so it's clear which benchmarks have which types

benchmarks/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+## Threading
+
+`__init__.py` pins `MKL_NUM_THREADS` to **4** when the machine has 4 or more
+physical cores, or falls back to **1** (single-threaded) otherwise. This keeps
+results comparable across CI machines in the shared pool regardless of their
+total core count. Physical cores are read from `/proc/cpuinfo` — hyperthreads
+are excluded per MKL recommendation.
+
+Override by setting `MKL_NUM_THREADS` in the environment before running ASV.
```
ndgrigorian: I would restructure this to focus first on how the user can override `MKL_NUM_THREADS` (i.e., "The number of threads can be set by setting `MKL_NUM_THREADS` in the environment [...]") and then discussing why we chose the defaults we did and what they are.

benchmarks/README.md
```diff
@@ -0,0 +1,85 @@
...
+## Running Locally
+
+> Benchmarks are designed for CI. Local runs require `mkl_fft` to be installed
+> in the active Python environment. Benchmarks that exercise SciPy interface
+> (`bench_scipy_fft.py`) also require SciPy:
+>
+> ```bash
+> python -m pip install -e ..
+> python -m pip install scipy
+> ```
```
ndgrigorian: don't think we need this, we can just refer to build and installation instructions

benchmarks/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+```bash
+cd benchmarks/
```
ndgrigorian: drop this, it directly contradicts the previous lines which assume the `benchmarks` folder is the cwd anyway, and the user should already understand it

benchmarks/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+This bypasses ASV environment management entirely — mkl_fft is pre-installed
+into a conda environment by the pipeline before ASV is invoked.
+
+- **Nightly (prod):** results are published to the benchmark dashboard
+- **PR (dev):** `asv compare` output is evaluated for regressions; a 30% slowdown
+  triggers a failed GitHub commit status
+
+Results are stored in the `mkl_fft-results` branch of
+`intel-innersource/libraries.python.intel.infrastructure.benchmark-dashboards`.
```
ndgrigorian: we don't need any of this and shouldn't reference innersource anywhere, external/open source users will just be confused

benchmarks/benchmarks/__init__.py
vchamarthi: They are irrelevant here. removed it so it does not signal users any relevance of OpenBLAS.

benchmarks/README.md
vchamarthi: That's overlooked,

benchmarks/README.md
vchamarthi: Makes sense, restructured it.

benchmarks/README.md
vchamarthi: Good point, Added everything to a table

benchmarks/README.md
vchamarthi: Refined the most part of the readme.

benchmarks/benchmarks/bench_fft1d.py
vchamarthi: restructured params and setup functions.

benchmarks/benchmarks/__init__.py
```diff
@@ -0,0 +1,48 @@
...
+    try:
+        with open("/proc/cpuinfo") as f:
+            content = f.read()
+        cpu_cores = int(re.search(r"cpu cores\s*:\s*(\d+)", content).group(1))
+        sockets = max(
+            len(set(re.findall(r"physical id\s*:\s*(\d+)", content))), 1
+        )
+        return cpu_cores * sockets
+    except Exception:
+        return 1
```
ndgrigorian: there is a small issue here: this restricts these benchmarks entirely to Linux, as the mechanism for getting the number of cores is Linux only

It may make more sense to use [psutil](https://pypi.org/project/psutil/) which handles platform logic, it seems to be the widely accepted solution for finding physical cores over logical cores

benchmarks/README.md
```diff
@@ -0,0 +1,29 @@
+# mkl_fft ASV Benchmarks
```
ndgrigorian, line 1: we should add a section on how a user can run these benchmarks, i.e., something which explains what is needed to be installed in an environment and what commands to run

benchmarks/benchmarks/__init__.py
```diff
@@ -0,0 +1,40 @@
...
+Strategy:
+  - Physical cores >= 4  →  fix MKL_NUM_THREADS = 4
+      4 is the lowest common denominator that guarantees multi-threaded MKL
+      behavior and is achievable on any modern CI machine.  Results from
+      different machines in the pool are therefore directly comparable.
+  - Physical cores < 4   →  fall back to MKL_NUM_THREADS = 1 (single-threaded)
+      Prevents over-subscription on under-resourced machines and avoids
+      misleading comparisons against 4-thread baselines.
+
+MKL recommendation: use physical cores, not logical (hyperthreaded) CPUs.
```
ndgrigorian: I think we should cut most of this, it isn't particularly relevant to this file as opposed to putting it in the README

It also isn't useful to external users to know the details of if we have dedicated runners or not, etc.

benchmarks/benchmarks/__init__.py
vchamarthi: good suggestion

benchmarks/README.md
vchamarthi: added

ndgrigorian (approved):
no other comments from my side, thank you @vchamarthi

vchamarthi:
@ndgrigorian  Thanks for the review, can you please merge it? i do not have enough permissions to merge.

ndgrigorian:
> @ndgrigorian Thanks for the review, can you please merge it? i do not have enough permissions to merge.

I'll merge tomorrow, after I give Anton a chance to take a quick look.

benchmarks/benchmarks/_utils.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def setup(self, shape, dtype):
+        rng = np.random.default_rng(_RNG_SEED)
+        cdtype = "complex64" if dtype == "float32" else "complex128"
+        if isinstance(shape, int):
+            half_shape = shape // 2 + 1
+        else:
+            half_shape = shape[:-1] + (shape[-1] // 2 + 1,)
+        self.x_real = rng.standard_normal(shape).astype(dtype)
+        self.x_complex = (
```
antonwolfy, line 67: That does not enforce Hermitian symmetry but is used for hfft benchmarks.

benchmarks/benchmarks/bench_numpy_fft.py
```diff
@@ -0,0 +1,157 @@
...
+_SIZES_1D = [256, 1024, 16384]
+
+
+# ---------------------------------------------------------------------------
+# 1-D complex-to-complex
+# ---------------------------------------------------------------------------
+
+
+class BenchC2C1D(BenchC2C):
```
antonwolfy: Would it make sense to combine benchmarks for numpy_fft and scipy_fft by adding module to the parameter list as scipy [did](https://github.com/scipy/scipy/blob/main/benchmarks/benchmarks/fft_basic.py#L86)?
The benchmark code seems similar for both numpy_fft and scipy_fft use cases.

benchmarks/benchmarks/_utils.py
```diff
@@ -0,0 +1,65 @@
...
+
+class BenchC2C:
+    """Base setup for complex-to-complex benchmarks.
+
+    Subclasses define params, param_names, and time_* / peakmem_* methods.
+    """
+
+    def setup(self, shape, dtype):
+        rng = np.random.default_rng(_RNG_SEED)
+        self.x = _make_input(rng, shape, dtype)
```
antonwolfy, line 30: I wonder if we need any warmup iteration. At the first run there will be time spent to create a DFTI descriptor handle. Each next run will reuse the same descriptor.
I wonder if it might be needed to consider that and to add a warmup iteration to exclude the first run from the measurements.

benchmarks/benchmarks/__init__.py
```diff
@@ -0,0 +1,21 @@
+"""ASV benchmarks for mkl_fft"""
+
+import os
```
antonwolfy, line 3: Would it make sense to add one more optional dependency declaration to `pyproject.toml`? Like: `benchmark = ["asv>=0.6", "psutil"]` or similar?

benchmarks/benchmarks/bench_fft1d.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+class BenchFFT1DNonPow2(BenchC2C):
+    """Forward and inverse complex FFT — non-power-of-two sizes.
+
+    MKL uses a different code path for non-power-of-two transforms;
+    this suite catches regressions in that path.
+    """
+
+    params = [_SIZES_NONPOW2, ["float64", "complex128", "complex64"]]
```
antonwolfy: Why we don't use `_DTYPES_REDUCED` or `_DTYPES_ALL` here?

.gitignore
```diff
@@ -9,3 +9,7 @@ mkl_fft/_pydfti.c
 mkl_fft/_pydfti.cpython*.so
 mkl_fft/_pydfti.*-win_amd64.pyd
 mkl_fft/src/mklfft.c
+
+# ASV benchmark artifacts
+.asv/
+benchmarks/.asv/
```
antonwolfy: The first pattern already covers the second one
```suggestion
.asv/
```

benchmarks/asv.conf.json
```diff
@@ -0,0 +1,19 @@
...
+    ],
+    "benchmark_dir": "benchmarks",
+    "env_dir": ".asv/env",
+    "results_dir": ".asv/results",
+    "html_dir": ".asv/html",
+    "build_cache_size": 2,
+    "default_benchmark_timeout": 500,
+    "regressions_thresholds": {
+        ".*": 0.3
+    }
```
antonwolfy, line 23: Do we need also to configure `environment_type`?

benchmarks/benchmarks/_utils.py
vchamarthi: Fixed

benchmarks/benchmarks/bench_numpy_fft.py
vchamarthi: Combined it as single interface that accepts modules

benchmarks/benchmarks/__init__.py
vchamarthi: Good suggestion, added it to optional deps

benchmarks/benchmarks/bench_fft1d.py
vchamarthi: parameterized this _DTYPES_ALL

benchmarks/asv.conf.json
vchamarthi: good point,
out target env is conda and also mentioned the channels.

benchmarks/benchmarks/_utils.py
vchamarthi: ASV's default `warmup_time` already cover this part, but I have added the explicit warmup call in, it just adds another bunch of lines to the code.

vchamarthi: locally ran with and without warmup 

  ┌─────────────────────┬─────────────────┬───────────────────┬───────┐
  │        Case         │ Cold first call │ Warm steady-state │ Ratio │
  ├─────────────────────┼─────────────────┼───────────────────┼───────┤
  │ fft n=1024 c128     │           263us │             2.5us │  105x │
  ├─────────────────────┼─────────────────┼───────────────────┼───────┤
  │ fft n=4096 c128     │           215us │             9.0us │   24x │
  ├─────────────────────┼─────────────────┼───────────────────┼───────┤
  │ rfft n=1024 f64     │           125us │             1.6us │   78x │
  ├─────────────────────┼─────────────────┼───────────────────┼───────┤
  │ fft2 (256,256) c128 │          42.4ms │            0.35ms │  122x │
  └─────────────────────┴─────────────────┴───────────────────┴───────┘

benchmarks/benchmarks/bench_fft1d.py
```diff
@@ -0,0 +1,111 @@
...
+
+    params = [_SIZES_POW2, _DTYPES_ALL]
+    param_names = ["n", "dtype"]
+
+    def setup(self, n, dtype):
+        super().setup(n, dtype)
+        # Prime the MKL DFTI descriptor cache so the first measured
+        # iteration doesn't pay the one-time descriptor-creation cost.
+        # ASV's warmup_time (default 0.1s) would normally cover this,
+        # but doing it explicitly removes the dependency on that default.
```
ndgrigorian:
```suggestion
        # prime MKL DFTI descriptor cache
```
comment is a bit verbose

benchmarks/benchmarks/_utils.py
```diff
@@ -0,0 +1,73 @@
...
+    if dt.kind == "c":
+        return (rng.standard_normal(s) + 1j * rng.standard_normal(s)).astype(dt)
+    return rng.standard_normal(s).astype(dt)
+
+
+class BenchC2C:
+    """Base setup for complex-to-complex benchmarks.
+
+    Subclasses define params, param_names, and time_* / peakmem_* methods.
+    Extra positional params (e.g. module) are accepted and ignored.
```
ndgrigorian: I think all cases of `accepted and ignored` in these comments should be changed, it's redundant and wordy.

`Other positional params are ignored' is good enough

benchmarks/benchmarks/_utils.py
```diff
@@ -0,0 +1,73 @@
...
+import numpy as np
+
+_RNG_SEED = 42
+
+
+def _make_input(rng, shape, dtype):
+    """Return an array of *shape* and *dtype*.
+
+    Complex dtypes get non-zero imaginary parts for a realistic signal.
+    *shape* may be an int (1-D) or a tuple.
```
ndgrigorian:
```suggestion
    `shape` may be an int (1-D) or a tuple.
```

benchmarks/benchmarks/bench_fft1d.py
```diff
@@ -0,0 +1,111 @@
...
+
+class BenchRFFT1D(BenchR2C):
+    """Forward rfft and inverse irfft — power-of-two sizes."""
+
+    params = [_SIZES_POW2, _DTYPES_REAL]
+    param_names = ["n", "dtype"]
+
+    def setup(self, n, dtype):
+        super().setup(n, dtype)
+        # Prime the DFTI descriptor cache (see BenchFFT1D.setup).
```
ndgrigorian: these comments can be removed: looking at BenchFFT1D's method already describes rationale, and that is naturally what a user will do if they wonder why setup is being called

benchmarks/benchmarks/bench_fftnd.py
```diff
@@ -0,0 +1,182 @@
...
+        _DTYPES_ALL,
+    ]
+    param_names = ["shape", "dtype"]
+
+    def setup(self, shape, dtype):
+        super().setup(shape, dtype)
+        # Prime the MKL DFTI descriptor cache so the first measured
+        # iteration doesn't pay the one-time descriptor-creation cost.
+        # ASV's warmup_time (default 0.1s) would normally cover this,
+        # but doing it explicitly removes the dependency on that default.
```
ndgrigorian: too wordy, same as the other

benchmarks/benchmarks/bench_fftnd.py
```diff
@@ -0,0 +1,182 @@
...
+            (50, 50, 50),
+            (30, 40, 50),  # non-cubic non-pow2
+        ],
+        _DTYPES_REDUCED,
+    ]
+    param_names = ["shape", "dtype"]
+
+    def setup(self, shape, dtype):
+        super().setup(shape, dtype)
+        # Prime the DFTI descriptor cache (see BenchFFT2D.setup).
```
ndgrigorian: I think these comments should be dropped too

benchmarks/benchmarks/bench_interfaces.py
```diff
@@ -0,0 +1,295 @@
...
+
+    def setup(self, n, dtype, module):
+        super().setup(n, dtype)
+        mod = _MODULE_MAP[module]
+        self.fft = mod.fft
+        self.ifft = mod.ifft
+        # Prime the MKL DFTI descriptor cache so the first measured
+        # iteration doesn't pay the one-time descriptor-creation cost.
+        # ASV's warmup_time (default 0.1s) would normally cover this,
+        # but doing it explicitly removes the dependency on that default.
```
ndgrigorian: same as the others

benchmarks/benchmarks/bench_interfaces.py
```diff
@@ -0,0 +1,295 @@
...
+
+    params = [_SIZES_1D, _DTYPES_REAL, _MODULES]
+    param_names = ["n", "dtype", "module"]
+
+    def setup(self, n, dtype, module):
+        super().setup(n, dtype)
+        mod = _MODULE_MAP[module]
+        self.rfft = mod.rfft
+        self.irfft = mod.irfft
+        # Prime the DFTI descriptor cache (see BenchC2C1D.setup).
```
ndgrigorian: can be dropped, same as the others

benchmarks/README.md
```diff
@@ -0,0 +1,67 @@
...
+## Notes on Measurement
+
+### DFTI descriptor warmup
+
+MKL creates a DFTI descriptor on the first FFT call for a given (size, dtype,
+strides) combination and reuses it on subsequent calls. To avoid charging
+that one-time cost to the first measured iteration, each benchmark's `setup`
+performs an explicit warmup call after preparing the input array. ASV's
+default `warmup_time` (0.1s) already amortizes this for sub-millisecond
+transforms, but the explicit warmup makes the intent visible.
```
ndgrigorian, line 40: I don't think this comment is necessary

benchmarks/benchmarks/bench_fft1d.py
vchamarthi: Gotcha

ndgrigorian (approved):
LGTM, will merge after CI run

benchmarks/benchmarks/bench_fft1d.py
```diff
@@ -0,0 +1,105 @@
...
+    """Forward and inverse complex FFT — power-of-two sizes."""
+
+    params = [_SIZES_POW2, _DTYPES_ALL]
+    param_names = ["n", "dtype"]
+
+    def setup(self, n, dtype):
+        super().setup(n, dtype)
+        # prime MKL DFTI descriptor cache
+        mkl_fft.fft(self.x)
+        mkl_fft.ifft(self.x)
```
ndgrigorian, line 26: what about adding a variable number of warmup runs to these setups, actually?

vchamarthi: I think one warmup run should be sufficient and ready with descriptor for steady state.

vchamarthi: Additionally ASV will by default perform some warmup runs apart from this explicit warm up.

ndgrigorian: if asv runs its own then it's fine

PR #319, opened 2026-05-08
task: module CLI patching methods

jharlow-intel:
adds `python -m mkl_fft patch install/uninstall/status` methods, and also `python -m mkl_fft with_patch` 
Convenience CLI functions that will make it way easier to let users have automagic mkl_fft optimizations in their workflows, such as if they want to run npbench suites but with mkl_fft optimizations.

Tested locally pretty exhaustively, everything worked pretty well. A lot of documentation is claude code and it was way too verbose. I deleted and edited a lot of it, but please take a look at it all and make sure it's correct.

If this gets merged, I'll do the same for mkl_random and mkl_umath

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,84 @@ numpy.allclose(mkl_res, np_res)
 ```
 
 ---
+# Patching Mechanisms
+
+`mkl_fft` provides convenient ways to enable MKL-accelerated FFT operations in NumPy with or without modifying your code. It supports both persistent patching (applies to all Python sessions) and one-shot execution (applies only to a single command). It also supports Python functions and context managers that do the same.
+
+## Persistent Patching
+
+### Install Persistent Patch
+
+```bash
+python -m mkl_fft patch install
+```
+
+### Check Patch Status
+
+```bash
+python -m mkl_fft patch status
+```
+
+Checks whether the persistent patch is currently installed. Returns exit code 0 if installed, 1 if not installed.
+
+### Uninstall Persistent Patch
+
+```bash
+python -m mkl_fft patch uninstall
+```
+
+Removes the persistent patch file, restoring NumPy to its default FFT implementation.
+
+## One-Shot Execution
+
+```bash
+python -m mkl_fft with_patch <command> [args...]
+```
+
+Runs a single command with MKL-accelerated FFT enabled. The patch is only active for that specific execution and does not persist.
+
+**Examples:**
+
+```bash
+# Run a Python script with MKL acceleration
+python -m mkl_fft with_patch python my_script.py
+
+# Run tests with MKL acceleration
+python -m mkl_fft with_patch python -m pytest tests/
+
+# Run a Python one-liner
+python -m mkl_fft with_patch python -c "import numpy; print(numpy.fft.fft.__module__)"
+
+# Run benchmarks with MKL acceleration
+python -m mkl_fft with_patch python run_benchmarks.py
+```
+
+## Programmatic Usage
+
+You can also patch NumPy programmatically in your Python code:
+
+```python
+import mkl_fft
+
+# Check if currently patched
+if mkl_fft.is_patched():
+    print("NumPy FFT is using MKL")
+
+# Enable patching globally
+mkl_fft.patch_numpy_fft()
+
+# Disable patching
+mkl_fft.restore_numpy_fft()
+
+# Use as context manager (recommended for temporary patching)
+with mkl_fft.mkl_fft():
+    # NumPy FFT uses MKL inside this block
+    import numpy as np
+    result = np.fft.fft(data)
+# NumPy FFT restored outside the block
+```
+
+---
 # Building from source
 
 To build `mkl_fft` from sources on Linux with Intel® oneMKL:
--- a/README.md
+++ b/README.md
@@ -84,79 +84,56 @@ numpy.allclose(mkl_res, np_res)
 ---
 # Patching Mechanisms
 
-`mkl_fft` provides convenient ways to enable MKL-accelerated FFT operations in NumPy with or without modifying your code. It supports both persistent patching (applies to all Python sessions) and one-shot execution (applies only to a single command). It also supports Python functions and context managers that do the same.
+`mkl_fft` provides convenient patch methods to enable MKL-accelerated FFT operations in NumPy with or without modifying your code.
 
-## Persistent Patching
+## CLI Quickstart
 
-### Install Persistent Patch
-
-```bash
-python -m mkl_fft patch install
-```
-
-### Check Patch Status
+### Persistent patch (all Python sessions)
 
 ```bash
-python -m mkl_fft patch status
+# Install
+python -m mkl_fft --patch install
+
+# Status (exit code: 0 = installed, 1 = not installed)
+python -m mkl_fft --patch status
+
+# Remove
+python -m mkl_fft --patch uninstall
 ```
 
-Checks whether the persistent patch is currently installed. Returns exit code 0 if installed, 1 if not installed.
-
-### Uninstall Persistent Patch
+### Verify current FFT backend
 
 ```bash
-python -m mkl_fft patch uninstall
+python -c "import numpy; print(f'numpy.fft.fft.__module__: {numpy.fft.fft.__module__}')"
 ```
 
-Removes the persistent patch file, restoring NumPy to its default FFT implementation.
-
-## One-Shot Execution
+### One-shot patch (single command only)
 
 ```bash
-python -m mkl_fft with_patch <command> [args...]
+# Script
+python -m mkl_fft --with-numpy-patch my_script.py
+
+# Pytest
+python -m mkl_fft --with-numpy-patch -m pytest tests/
+
+# One-liner
+python -m mkl_fft --with-numpy-patch -c "import numpy; print(f\"numpy.fft.fft.__module__: {numpy.fft.fft.__module__}\")"
+
+# Non-Python command
+python -m mkl_fft --with-numpy-patch -- <command> [args...]
 ```
 
-Runs a single command with MKL-accelerated FFT enabled. The patch is only active for that specific execution and does not persist.
-
-**Examples:**
-
-```bash
-# Run a Python script with MKL acceleration
-python -m mkl_fft with_patch python my_script.py
-
-# Run tests with MKL acceleration
-python -m mkl_fft with_patch python -m pytest tests/
-
-# Run a Python one-liner
-python -m mkl_fft with_patch python -c "import numpy; print(numpy.fft.fft.__module__)"
-
-# Run benchmarks with MKL acceleration
-python -m mkl_fft with_patch python run_benchmarks.py
-```
-
-## Programmatic Usage
-
-You can also patch NumPy programmatically in your Python code:
+## Programmatic Quickstart
 
 ```python
 import mkl_fft
 
-# Check if currently patched
-if mkl_fft.is_patched():
-    print("NumPy FFT is using MKL")
-
-# Enable patching globally
 mkl_fft.patch_numpy_fft()
-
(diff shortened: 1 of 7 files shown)


mkl_fft/with_patch.py
```diff
@@ -0,0 +1,92 @@
...
+        print("  python -m mkl_fft with_patch python -c 'import numpy; print(numpy.fft.fft.__module__)'")
+        sys.exit(1)
+
+    args = parsed_args.command
+
+    patch_script = "import mkl_fft; mkl_fft.patch_numpy_fft()"
+
+    env = os.environ.copy()
+
+    if "PYTHONSTARTUP" in env:
```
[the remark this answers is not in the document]
jharlow-intel: That's actually a good point

jharlow-intel:
@antonwolfy @ndgrigorian ready for review, I'll open this for the other 2 mkl packages once this gets merged.

Tried to get thorough CLI testing added to the meta.yaml, but conda seems to do something weird with the PYTHONPATH while running the tests so I had to run a lot of integration tests by hand

mkl_fft/__main__.py
```diff
@@ -0,0 +1,67 @@
...
+# FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+# DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+# SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+# OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+"""Command-line interface for mkl_fft."""
+
+import sys
```
ndgrigorian, line 29: I think we should use argparse here, dpctl is a good example
https://github.com/IntelPython/dpctl/blob/master/dpctl/__main__.py

it's a bit cleaner

mkl_fft/patch.py
```diff
@@ -0,0 +1,138 @@
...
+
+PTH_CONTENT = """import mkl_fft._patch_startup"""
+
+
+def install_patch():
+    """Install persistent NumPy FFT patch using .pth file."""
+    pth_path = get_pth_path()
+
+    if pth_path.exists():
+        print(f"Persistent patch already installed at {pth_path}")
```
antonwolfy: I guess it would be proper to have that as a warning, rather then default trace to the console.

mkl_fft/patch.py
```diff
@@ -0,0 +1,138 @@
...
+        return
+
+    try:
+        pth_path.parent.mkdir(parents=True, exist_ok=True)
+        pth_path.write_text(PTH_CONTENT)
+        print(f"Persistent patch installed at {pth_path}")
+        print()
+        print("NumPy FFT will now use MKL-accelerated implementations in all")
+        print("Python sessions. To disable, run:")
+        print("  python -m mkl_fft patch uninstall")
```
antonwolfy: I believe it should be only traced under verbose mode

mkl_fft/tests/test_cli.py
```diff
@@ -0,0 +1,92 @@
...
+    assert "import mkl_fft._patch_startup" in content
+
+    captured = capsys.readouterr()
+    assert "Persistent patch installed" in captured.out
+
+
+def test_install_patch_already_installed(mock_pth_path, capsys):
+    """Test installing patch when already installed."""
+    install_patch()
+    install_patch()
```
antonwolfy: It should be warning raised at second installation and then we will verify that here through appropriate pytest i/f.

mkl_fft/__main__.py
antonwolfy: Agree, then it would be something like: `python -m mkl_fft --patch [args]`

mkl_fft/__main__.py
```diff
@@ -0,0 +1,67 @@
...
+        print()
+        print("Commands:")
+        print("  patch install      Install persistent NumPy FFT patch")
+        print("  patch uninstall    Uninstall persistent NumPy FFT patch")
+        print("  patch status       Check if persistent patch is installed")
+        print("  with_patch <cmd>   Run command with temporary NumPy FFT patch")
+        print()
+        print("Examples:")
+        print("  python -m mkl_fft patch install")
+        print("  python -m mkl_fft with_patch python script.py")
```
antonwolfy: Why do we need the second `python` in the command? Is there any use case when we will run something different from python file?

mkl_fft/tests/test_cli.py
```diff
@@ -0,0 +1,92 @@
...
+
+    captured = capsys.readouterr()
+    assert "already installed" in captured.out
```
antonwolfy: Do we have a better way to check the patch is enabled through importing mkl_fft and checking `mkl_fft.is_patched()` result?

mkl_fft/_patch_startup.py
```diff
@@ -0,0 +1,33 @@
+# Copyright (c) 2017, Intel Corporation
```
antonwolfy: Should we use the current year for new files?
```suggestion
# Copyright (c) 2026, Intel Corporation
```

mkl_fft/patch.py
```diff
@@ -0,0 +1,138 @@
...
+    except OSError as e:
+        print(f"Error installing patch: {e}")
```
antonwolfy: Would it be better to handle through raising an exception?

mkl_fft/__main__.py
antonwolfy: It might be helpful also to support something like:
`python -m mkl_fft with_patch "import numpy as np; ...; res = np.fft.fft(data); ..."`

mkl_fft/patch.py
ndgrigorian: agreed, that does make more sense

mkl_fft/_patch_startup.py
jharlow-intel: Agreed, fixed

mkl_fft/patch.py
jharlow-intel: Good point.

I've added a `verbose=False` var, that the CLI can set to true by passing `python -m mkl_fft -v/--verbose` now.

mkl_fft/patch.py
jharlow-intel: Yeah, that would be better! I'll make some custom exception classes if we need any and replace all the sys.exit(1) with exception raises

mkl_fft/patch.py
jharlow-intel: Also agreed!

mkl_fft/__main__.py
jharlow-intel: Fixed, used argparse here and use the pattern you suggested

mkl_fft/__main__.py
jharlow-intel: Yeah I honestly should not have let that through, I've modified it to work as you suggsted, with a simple `python -m mkl_fft --with-patch -c "print('hello world')"` type commands as well as just running basic python scripts like `python -m mkl_fft --with-patch script.py`

mkl_fft/tests/test_cli.py
jharlow-intel: Yes, I've added an additional test that will do the .pth patching, and then use `is_patched` to check

README.md
```diff
@@ -81,6 +81,84 @@ numpy.allclose(mkl_res, np_res)
 # True
 ```
 
+---
+# Patching Mechanisms
+
+`mkl_fft` provides convenient ways to enable MKL-accelerated FFT operations in NumPy with or without modifying your code. It supports both persistent patching (applies to all Python sessions) and one-shot execution (applies only to a single command). It also supports Python functions and context managers that do the same.
```
ndgrigorian:
```suggestion
`mkl_fft` provides patches to enable MKL-accelerated FFT operations in NumPy with or without modifying your code. It supports both persistent patching (applies to all Python sessions) and one-shot execution (applies only to a single command). It also supports Python functions and context managers that do the same.
```

mkl_fft/with_patch.py
```diff
@@ -0,0 +1,97 @@
...
+
+"""Run Python commands with temporary NumPy FFT patch."""
+
+import os
+import subprocess
+import sys
+import tempfile
+
+
+def run_with_patch(args):
```
ndgrigorian: I think we should rename "with_patch" in general to "patch_numpy" in case we for some reason, were to someday want to patch other libraries (e.g., SciPy)

jharlow-intel: Done! Added `with_patch_numpy` method and `--with-patch-numpy` CLI arg

mkl_fft/with_patch.py
```diff
@@ -0,0 +1,99 @@
...
+        try:
+            os.unlink(sitecustomize_path)
+            os.rmdir(temp_dir)
+        except OSError:
+            pass
+
+
+def main(args=None):
+    """Deprecated entry point. Use run_with_numpy_patch() instead."""
+    run_with_numpy_patch(args if args else sys.argv[1:])
```
ndgrigorian: why do we have this deprecated entry point? This is new functionality so, why do we need deprecation?

mkl_fft/with_patch.py
```diff
@@ -0,0 +1,99 @@
...
+        print("  python -m mkl_fft --with-numpy-patch -m pytest tests/")
+        print("  python -m mkl_fft --with-numpy-patch -- any command here")
+        sys.exit(1)
+
+    # If args start with '--', strip it and run as-is (arbitrary command)
+    if args[0] == "--":
+        args = args[1:]
+    # Otherwise, prepend 'python' for convenience
+    else:
+        args = [sys.executable] + args
```
ndgrigorian, line 58: since we're passing `args.with_numpy_patch` into `run_with_numpy_patch` (which should just be the command) do we need this strip?

ndgrigorian: oh I see the reason for this now, nevermind

mkl_fft/with_patch.py
```diff
@@ -0,0 +1,99 @@
...
+        print("Examples:")
+        print("  python -m mkl_fft --with-numpy-patch script.py")
+        print(
+            "  python -m mkl_fft --with-numpy-patch -c 'import numpy; print(numpy.fft.fft.__module__)'"
+        )
+        print("  python -m mkl_fft --with-numpy-patch -m pytest tests/")
+        print("  python -m mkl_fft --with-numpy-patch -- any command here")
+        sys.exit(1)
+
+    # If args start with '--', strip it and run as-is (arbitrary command)
```
ndgrigorian:
```suggestion
    # strip leading '--' and run as-is
```

mkl_fft/with_patch.py
```diff
@@ -0,0 +1,99 @@
...
+    if args[0] == "--":
+        args = args[1:]
+    # Otherwise, prepend 'python' for convenience
```
ndgrigorian:
```suggestion
    # prepend 'python' for convenience
```

mkl_fft/with_patch.py
jharlow-intel: Oops, leftover from one of the refactors that I overlooked - everything is using `__main__`, so I'll just delete this `main()` function then.

ndgrigorian (approved):

ndgrigorian:
@jharlow-intel 
thanks for adding this. LGTM, merging

PR #326, opened 2026-05-12
Fix conda-package workflow

ndgrigorian:
with NumPy no longer on Intel channel, conda-package needs the same work-around as conda-package-cf


antonwolfy (approved):
Thank you @ndgrigorian for recovering the CI

PR #331, opened 2026-06-05
Add requirements.txt, missing patch to ASV benchmarks

vchamarthi:
- Add `_patch_setup.py` that calls `patch_numpy_fft()` at worker import time with hard-fail validation (ensures benchmarks never silently run with stock NumPy)
- Add `requirements.txt` with `psutil` and `scipy` benchmark dependencies
- Update the regression threshold matching the mkl_umath


benchmarks/requirements.txt
```diff
@@ -0,0 +1,2 @@
+psutil
+scipy
```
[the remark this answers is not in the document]
vchamarthi: Jenkins CI benchmarks running script know which recipe it is benchmarking, so not needed.

benchmarks/benchmarks/_patch_setup.py
```diff
@@ -0,0 +1,65 @@
+"""MKL patch setup — executed once per ASV worker process at import time.
+
+Patches NumPy FFT with the Intel MKL FFT implementation.
+Hard-fails with a descriptive RuntimeError if mkl_fft is missing or the
+patch does not take effect, so benchmarks never silently run on stock NumPy.
+"""
```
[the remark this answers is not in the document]
vchamarthi: Intentional

benchmarks/benchmarks/_patch_setup.py
```diff
@@ -0,0 +1,65 @@
...
+        "mkl_fft": lambda: np.fft.fft.__module__,
+    }
+    for mod_name in patched:
+        try:
+            attr = _attr_checks[mod_name]()
+        except Exception:
+            attr = "unknown"
+        print(f"[mkl-patch] {mod_name}: numpy.fft dispatch -> {attr}")
+
+    print("[mkl-patch] ALL OK -- mkl_fft active")
```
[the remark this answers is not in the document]
vchamarthi: These prints run once per ASV worker process during setup, wouldn't be a noisy log.
If code reaches this print statement it should be patches successfully otherwise it would raise exception.

benchmarks/benchmarks/_patch_setup.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+def _apply_patches():
+    import numpy as np
+
+    patched = {}
+
+    for mod_name, patch_fn_name in _PATCH_MAP:
+        try:
+            mod = __import__(mod_name)
```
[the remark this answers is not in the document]
vchamarthi: Speculative future proofing...

benchmarks/benchmarks/_patch_setup.py
```diff
@@ -0,0 +1,67 @@
+"""MKL patch setup — executed once per ASV worker process at import time.
+
+Patches NumPy FFT with the Intel MKL FFT implementation.
+Hard-fails with a descriptive RuntimeError if mkl_fft is missing or the
+patch does not take effect, so benchmarks never silently run on stock NumPy.
+"""
+
+_PATCH_MAP = [
+    ("mkl_fft", "patch_numpy_fft"),
+]
```
ndgrigorian: making this a dictionary would make more sense to me, especially with the name "PATCH_MAP"

benchmarks/benchmarks/_patch_setup.py
```diff
@@ -0,0 +1,67 @@
...
+                f"  Ensure the conda env contains {mod_name} "
+                f"from the Intel channel.\n"
+                "  Required channels: "
+                "https://software.repos.intel.com/python/conda"
+            ) from exc
+
+        patch_fn = getattr(mod, patch_fn_name, None)
+        if patch_fn is None:
+            raise RuntimeError(
+                f"[mkl-patch] {mod_name} has no {patch_fn_name}(). "
```
ndgrigorian:
```suggestion
                f"{mod_name} has no {patch_fn_name}(). "
```

ndgrigorian: all `[mkl-patch]` should be removed I feel, it's a bit confusing, makes it seem like a package (we have `mkl-service` after all).

benchmarks/benchmarks/_patch_setup.py
```diff
@@ -0,0 +1,67 @@
...
+        "mkl_fft": lambda: np.fft.fft.__module__,
+    }
+    for mod_name in patched:
+        try:
+            attr = _attr_checks[mod_name]()
+        except Exception:
+            attr = "unknown"
+        print(f"[mkl-patch] {mod_name}: numpy.fft dispatch -> {attr}")
+
+    print("[mkl-patch] ALL OK -- mkl_fft active")
```
ndgrigorian: I think we can drop this line -- the patching already has a verbose mode. If anything, we could just call into the verbose mode, I feel.

benchmarks/benchmarks/_patch_setup.py
```diff
@@ -0,0 +1,67 @@
...
+
+    _attr_checks = {
+        "mkl_fft": lambda: np.fft.fft.__module__,
+    }
+    for mod_name in patched:
+        try:
+            attr = _attr_checks[mod_name]()
+        except Exception:
+            attr = "unknown"
+        print(f"[mkl-patch] {mod_name}: numpy.fft dispatch -> {attr}")
```
ndgrigorian: same here, a call with verbose patching could substitute

benchmarks/requirements.txt
```diff
@@ -0,0 +1,2 @@
+psutil
```
antonwolfy, line 1: Why do we need that if we have the appropriate optional dep in `pyproject.toml` already?
Otherwise it seems they have to be aligned, but now it's not the case.

benchmarks/benchmarks/_patch_setup.py
```diff
@@ -0,0 +1,64 @@
+"""MKL patch setup — executed once per ASV worker process at import time.
+
+Patches NumPy FFT with the Intel MKL FFT implementation.
+Hard-fails with a descriptive RuntimeError if mkl_fft is missing or the
+patch does not take effect, so benchmarks never silently run on stock NumPy.
+"""
+
+PATCH_MAP = {
+    "mkl_fft": "patch_numpy_fft",
```
antonwolfy: I failed to find any benchmark which calls `np.fft.*`. Why do we need to enable the NumPy FFT patch?

benchmarks/benchmarks/_patch_setup.py
```diff
@@ -0,0 +1,64 @@
...
+            )
+
+        patched[mod_name] = mod
+
+    _attr_checks = {
+        "mkl_fft": lambda: np.fft.fft.__module__,
+    }
+    for mod_name in patched:
+        try:
+            _attr_checks[mod_name]()
```
antonwolfy: Why do we need that? In any negative case we will just silently pass.

benchmarks/requirements.txt
antonwolfy: In `README.md` we have: `pip install asv psutil`. It sounds incomplete. Should we use `pip install -e .[benchmark]` instead?

benchmarks/requirements.txt
antonwolfy: Do we need to add `scipy>=1.10` to the benchmark opt dep?

vchamarthi: yes, added scipy to benchmark opt dep

benchmarks/benchmarks/_patch_setup.py
vchamarthi: You're right, removed. All benchmarks call mkl_fft.* directly.

We do not need that since we are timing mkl_fft interfaces directly.

benchmarks/requirements.txt
vchamarthi: I updated the README.md

requirements.txt is used by the Jenkins CI script (benchmark_run.sh), which installs deps via conda install --file requirements.txt — it can't use pip install -e ".[benchmark]" because the env installs mkl_fft from the built conda package, not from source. The two files serve different package managers: requirements.txt for CI, pyproject.toml for local dev.

benchmarks/benchmarks/_patch_setup.py
vchamarthi: removed along with _patch_setup.py

benchmarks/requirements.txt
antonwolfy: It might be needed to pin as `scipy>=1.10` to be aligned with pyproject.toml

antonwolfy (approved):
Thank you @vchamarthi, only small nit from me.

benchmarks/README.md
```diff
@@ -44,7 +44,7 @@ transforms, but the explicit warmup makes the intent visible.
 Prerequisites:
 
 ```bash
-pip install asv psutil
+pip install -e ".[benchmark]"
```
antonwolfy: probably we shouldn't use `-e` here, to claim install the tooling rather the full compile from the source, since otherwise the editable install requires the full build chain (Cython + mkl-devel)
```suggestion
pip install ".[benchmark]"
```

vchamarthi: Good suggestion :)

antonwolfy (approved):

PR #349, opened 2026-07-30
Prepare for 2.3.0 release

vlad-perevezentsev:
The PR updates `__version__` and CHANGELOG.md to publish the new 2.3.0 release


CHANGELOG.md
```diff
@@ -4,15 +4,14 @@ All notable changes to this project will be documented in this file.
...
+## [2.3.0] - 2026-07-30
 
 ### Added
 * Added CLI for persistent and ephemeral NumPy FFT patching: `python -m mkl_fft patch install/uninstall/status` for persistent patching across all Python sessions, and `python -m mkl_fft with_patch <command>` for one-shot execution with MKL acceleration [gh-319](https://github.com/IntelPython/mkl_fft/pull/319)
 
 ### Changed
 * Removed `numpy-base` dependency and `USE_NUMPY_BASE` environment variable from conda recipe [gh-318](https://github.com/IntelPython/mkl_fft/pull/318)
-
-### Fixed
+* Migrated the build system from `setuptools` to `meson-python`, removing `setup.py` in favor of `meson.build` [gh-303](https://github.com/IntelPython/mkl_fft/pull/303)
```
antonwolfy, line 14: It'd make sense to state also that the PR also moves to mkl as dependency, with mkl-service only used as a dependency when using the scipy interface. If scipy is installed without mkl-service, the scipy interface won't be available.
Or to add that by separate changelog entry.

antonwolfy (approved):
The only small nit from me

CHANGELOG.md
vlad-perevezentsev: @antonwolfy this is not quite accurate for #303 because this PR is only the build-system migration and does not update runtime deps. 
That behavior was introduced earlier: mkl-service made optional in #305 (2.2.0)
mkl dropped as a runtime dep in #322 (2.2.1) and both already in the `CHANGELOG.md` 
I will remove this mention from the #303 description.

PR #351, opened 2026-07-31
Apply RPATH logic for OSX

ndgrigorian:
This PR fixes an oversight when mkl_fft was updated to use meson-python, namely, that mkl_fft is built for OSX on conda-forge. Logic for the RPATH was removed.

This PR reverts that change, aligning to mkl-service


vlad-perevezentsev (approved):
LGTM
Thank you @ndgrigorian !

PR #358, opened 2026-08-03
Fix macOS rpath: emit one -rpath per path instead of colon-joining

vlad-perevezentsev:
In #351 the extensions' runtime search path by colon-joining two paths into a single -Wl,-rpath, argument:
```
-Wl,-rpath,$ORIGIN/../..:$ORIGIN/../../..
```

Colon-as-separator is an **ELF/Linux** convention. On macOS (Mach-O) each `-rpath` produces exactly one `LC_RPATH` load command and the string is taken **verbatim** — the colon is not a separator. As a result, macOS extensions shipped a single unusable rpath entry:
```
LC_RPATH: @loader_path/../..:@loader_path/../../..   ← no such directory
```

This PR emits one `-Wl,-rpath,` argument per path, restoring the exact per-entry behavior of the legacy `setup.py` (`runtime_library_dirs=["$ORIGIN/../..", "$ORIGIN/../../.."]`). This is correct on both Mach-O and ELF.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -35,8 +35,11 @@ if host_machine.system() != 'windows'
     else
         origin = '$ORIGIN'
     endif
-    rpath = origin / '../..' + ':' + origin / '../../..'
-    rpath_link_args = ['-Wl,-rpath,' + rpath]
+
+    rpath_link_args = [
+        '-Wl,-rpath,' + origin / '../..',
+        '-Wl,-rpath,' + origin / '../../..',
+    ]
 endif
 
 mkl_dep = dependency('MKL', method: 'cmake',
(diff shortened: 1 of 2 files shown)


meson.build
```diff
@@ -35,8 +35,11 @@ if host_machine.system() != 'windows'
     else
         origin = '$ORIGIN'
     endif
-    rpath = origin / '../..' + ':' + origin / '../../..'
-    rpath_link_args = ['-Wl,-rpath,' + rpath]
+
```
antonwolfy, line 38: Missed to update the changelog

ndgrigorian (approved):

PR #360, opened 2026-08-04
Group Dependabot version updates into a single weekly PR per ecosystem

antonwolfy:
This PR extends the Dependabot configuration so pre-commit hook updates are tracked and bumped automatically, and consolidates version bumps into a single weekly PR per ecosystem instead of one PR per bump.

Previously only the `github-actions` ecosystem was covered, and each bump landed in its own PR.

## Changes

- **`github-actions`** — adds a `groups` block with `patterns: ["*"]` so all action bumps are consolidated into a single weekly PR, and applies the `autoupdate` label.
- **`pre-commit`** (`/`) — new ecosystem entry that updates hook versions in `.pre-commit-config.yaml`, grouped into a single weekly PR and labeled.

Also deletes `.github/workflows/pre-commit-autoupdate.yml`. Its flow is now handled by the `pre-commit` Dependabot entry.


vlad-perevezentsev (approved):
LGTM
Thank you @antonwolfy

PR #370, opened 2026-08-26
fix: norm scaling

### LINKED ISSUE #336: norm="ortho" does not take into account the actual transformed axes
Using `norm="ortho"` the array norm should be preserved but it is not:
``` python
import numpy as np
import mkl_fft

s = 16,18
ax = 0,

a = np.random.uniform(0,1,s) + 1j * np.random.uniform(0,1,s)
a /= np.sqrt((abs(a)**2).sum()) # normalise

a_np = np.fft.fftn(a, axes=ax, norm="ortho")
a_mkl = mkl_fft.fftn(a, axes=ax, norm="ortho")

print(f"  Array norm: {(abs(a)**2).sum():.6f}")
print(f" np-fft norm: {(abs(a_np)**2).sum():.6f}")
print(f"mkl_fft norm: {(abs(a_mkl)**2).sum():.6f}")
```

Gives:
```
  Array norm: 1.000000
 np-fft norm: 1.000000
mkl_fft norm: 0.055556
```

The ratio of the norms here is 18, the size of the non-transformed axis. So `norm="ortho"` does not take into account the actual list of transformed axes.

Tested with `numpy 2.3.5` and `mkl_fft 2.2.1`, all from conda-forge.

jharlow-intel:
Doing some iterative agentic looping dev experimentation. Part of it found this:

Two `norm` scaling bugs in the root N-D API returned silently mis-scaled
results. Both are fixed by resolving the scale basis from the *transformed*
axes, matching what `mkl_fft.interfaces.*` already does via `_cook_nd_args`.

1. **Subset-axes transforms over-normalized** by the product of the
   untransformed axis lengths, because the scale was computed over the full
   array shape. Hits `fftn(x, axes=(0,))`, and less obviously `fft2(x)` on a
   3-D array — that transforms 2 of 3 axes.
2. **`irfftn`/`irfft2` normalized over the input length** `n` rather than the
   complex-to-real output length `2 * (n - 1)` along the last transformed
   axis. Wrong even when every axis was transformed.

Applies to `fftn`/`ifftn`/`rfftn`/`irfftn` and the `fft2`/`ifft2`/`rfft2`/
`irfft2` family with `norm="forward"` or `"ortho"` and no explicit `s`.
Unaffected: `norm=None`/`"backward"`, explicit `s=`, 1-D transforms, and
`interfaces.numpy_fft`/`scipy_fft`.

## Why it wasn't caught

The existing N-D norm tests compare `mkl_fft` against other `mkl_fft` calls,
and `test_fft_with_order` compares it against itself across memory layouts —
self-consistency, never an external reference. The new
`test_dispatch_equivalence.py` uses `numpy.fft` as the reference across
dtype × layout × axes × norm, on a shape whose axis lengths all differ so that
an axis permutation cannot produce a correctly shaped result.

## Testing

1725 passed / 104 skipped (existing suite was 971 — no regressions).
176 root-API combinations checked against `numpy.fft`: 0 mismatches, 32 of
them failing before the fix. The `norm=None` path is unchanged; the new helper
short-circuits in ~0.04 µs.

This is part of other on-going performance improvement looping I'm doing. No rush in merging it, it was entirely agentic and I didn't have time to thoroughly review, hopefully an expert here can say whether the PR is correct or not

Closes #336


jharlow-intel:
reproducer:
```
conda create -n mkl_fft_test --override-channels -c https://software.repos.intel.com/python/conda -c conda-forge python mkl_fft

conda activate mkl_fft_test

# python execute following script
```

```python
import sys

import numpy as np

import mkl_fft

x = np.random.default_rng(0).standard_normal((8, 7, 13)) + 0j
bad = 0


def check(label, got, want):
    global bad
    f = np.vdot(want, got) / np.vdot(want, want)  # least-squares scale
    pure = np.allclose(got, f * want)  # wrong by ONLY that scale?
    ok = abs(f - 1) < 1e-9 and pure
    bad += not ok
    note = "" if pure else "   <-- not a pure scale, values differ too!"
    print(f"  {'ok ' if ok else 'BUG'}  {label:<38} scale={f.real:9.6f}{note}")


print(f"mkl_fft {mkl_fft.__version__}, numpy {np.__version__}, x.shape={x.shape}\n")

print("subset of axes, s not given:")
for axes in [(0,), (1,), (2,), (1, 2)]:
    for norm in ("forward", "ortho"):
        check(
            f"fftn(axes={axes}, norm={norm!r})",
            mkl_fft.fftn(x, axes=axes, norm=norm),
            np.fft.fftn(x, axes=axes, norm=norm),
        )

print("fft2 on a 3-D array -- transforms 2 of 3 axes:")
for norm in ("forward", "ortho"):
    check(
        f"fft2(norm={norm!r})",
        mkl_fft.fft2(x, norm=norm),
        np.fft.fft2(x, norm=norm),
    )

print("complex-to-real, every axis transformed:")
for fn in ("irfftn", "irfft2"):
    for norm in ("forward", "ortho"):
        check(
            f"{fn}(norm={norm!r})",
            getattr(mkl_fft, fn)(x, norm=norm),
            getattr(np.fft, fn)(x, norm=norm),
        )

print("\ncontrols that should always pass:")
check("fftn(axes=None, norm='ortho')", mkl_fft.fftn(x, norm="ortho"), np.fft.fftn(x, norm="ortho"))
check("fft(axis=1, norm='ortho')", mkl_fft.fft(x, axis=1, norm="ortho"), np.fft.fft(x, axis=1, norm="ortho"))
check("fftn(axes=(0,), norm=None)", mkl_fft.fftn(x, axes=(0,)), np.fft.fftn(x, axes=(0,)))

print(f"\n{bad} mismatched -> bug present" if bad else "\nall match -> fixed")
sys.exit(1 if bad else 0)
```

ndgrigorian:
@jharlow-intel can you check if this also covers https://github.com/IntelPython/mkl_fft/issues/336?

mkl_fft/tests/test_dispatch_equivalence.py
```diff
@@ -0,0 +1,230 @@
+"""Cross-library equivalence checks for axis and axes dispatch.
```
antonwolfy, line 26: Do we need to add copyright header?

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 ### Fixed
+* Fixed `norm="forward"` and `norm="ortho"` scaling in `mkl_fft.fftn`, `ifftn`, `rfftn`, `irfftn` and the `fft2`/`ifft2`/`rfft2`/`irfft2` family when only a subset of the input axes is transformed and `s` is not given. The scale factor was computed over the full array shape instead of over the transformed axes, over-normalizing the result by the product of the untransformed axis lengths (for example `fft2` on a 3-D array, or `fftn(x, axes=(0,))`). The `mkl_fft.interfaces.numpy_fft` and `mkl_fft.interfaces.scipy_fft` wrappers were unaffected, as they resolve `s` before delegating
```
antonwolfy: Links to PR are missing

mkl_fft/_fft_utils.py
```diff
@@ -78,6 +78,43 @@ def _compute_fwd_scale(norm, n, shape):
...
+    ``_compute_fwd_scale``, which validates them.
+    """
+
+    if s is not None or norm in (None, "backward"):
+        return s
+    try:
+        if axes is None:
+            ss = list(x.shape)
+            last = len(ss) - 1
+        else:
```
antonwolfy, line 101: Empty axis set → scale 1.0, matches numpy:
```suggestion
        elif len(axes) == 0:
            return ()
        else:
```
It probably would be good to cover that in tests

mkl_fft/tests/test_dispatch_equivalence.py
```diff
@@ -0,0 +1,230 @@
...
+@pytest.mark.parametrize("norm", [None, "backward", "forward", "ortho"])
+def test_fft_axis_norm(func, dtype, axis, norm):
+    x = _make(_SHAPE_3D, dtype)
+    got = getattr(mkl_fft, func)(x, axis=axis, norm=norm)
+    want = getattr(np.fft, func)(x, axis=axis, norm=norm)
+    _check(got, want, dtype)
+
+
+# ---------------------------------------------------------------------------
+# out= must not change results on any dispatch path
```
antonwolfy, line 270: Nice fix — the norm×subset-axes and c2r output-length bugs are correctly resolved, and the new `test_dispatch_equivalence.py` fails loudly on master (74 failures) and passes cleanly after. Two optional follow-ups on test coverage; I verified every case below **passes against the fixed backend**, so these are regression guards for untested-but-correct paths, not new bugs.

### Gap A — `out=` coverage

The only `out=` test (`test_fftn_axes_subset_out`) covers `fftn` / complex only / **no `norm`**.

| Missing case | Why it matters | Example |
|---|---|---|
| `out=` + scaled `norm=` | The PR's whole subject — the scale is written *into* `out`; nothing checks the scaled-write path | `mkl_fft.fftn(x, axes=(0,), norm="forward", out=out)` |
| `out=` on `ifftn` | Shares the `out` path but never exercised with it | `mkl_fft.ifftn(x, out=out)` |
| `out=` on `rfftn` | r2c: `out` is complex with reduced last axis `n//2+1` — different allocation than c2c | `out = np.empty((8,7,7), complex128); mkl_fft.rfftn(xr, out=out)` |
| `out=` on `irfftn` (+ norm) | c2r: `out` is **real** with the expanded `2*(n-1)` last axis — exactly the length the `invreal` branch computes | `out = np.empty((8,7,24), float64); mkl_fft.irfftn(xc, norm="forward", out=out)` |

### Gap B — `s=` combined with a scaled `norm`

The helper deliberately early-returns when `s is not None`, so `_compute_fwd_scale` uses `prod(s)`. That branch is only tested without norm; the `s=` × scaled-norm intersection is untested.

| Missing case | Why it matters | Example |
|---|---|---|
| `s=` pads a subset axis + `forward`/`ortho` | Scale must be `prod(s)` (padded length), not the original axis length | `mkl_fft.fftn(x, s=(16,), axes=(0,), norm="forward")` |
| `s=` truncates + scaled norm | Same, downward | `mkl_fft.fftn(x, s=(4,), axes=(0,), norm="ortho")` |
| **`irfftn` with explicit `s` + scaled norm** | When `s` is given, the `invreal` doubling `2*(n-1)` must **not** apply — a regression that double-applies it only surfaces here | `mkl_fft.irfftn(xc, s=(8,7,20), norm="forward")` → normalizes over `20`, not `2*(13-1)` |
| `rfftn` with `s=` on a subset + scaled norm | Locks the r2c `s`-given path with scaling | `mkl_fft.rfftn(xr, s=(10,), axes=(1,), norm="ortho")` |

CHANGELOG.md
jharlow-intel: fixed, and shortened changelog entry

mkl_fft/tests/test_dispatch_equivalence.py
jharlow-intel: added the suggested additional test coverage

jharlow-intel:
@ndgrigorian this seems to indeed fix the issue detailed in https://github.com/IntelPython/mkl_fft/issues/336

jharlow-intel:
@ndgrigorian  @vlad-perevezentsev  added some tests to cover what that user ran into. This is probably ready for a review from the experts
CI failures were due to internal channel issue, they should work on next run

mkl_fft/tests/test_dispatch_equivalence.py
```diff
@@ -0,0 +1,422 @@
...
+
+These cover the paths that dispatch on *which* axes are requested: full-axes
+transforms reach the batched N-D descriptor, strict subsets iterate the
+complementary axes, and 1-D transforms of rank > 2 arrays are batched only for
+the first and last axis.
+"""
+
+import itertools
+
+import numpy as np
```
vlad-perevezentsev, line 53: It looks like `numpy.fft` may be patched to `mkl_fft` when NumPy is installed from Intel channels (software.repos.intel.com/python/conda or gold)
 ```
$ python -c "import numpy as np; print(np.fft.fftn.__module__)"
mkl_fft.interfaces._numpy_fft
```
This means the tests may not always use the real `numpy.fft` implementation as the reference, contrary to the docstring on 35 line

Could we use the NumPy implementation directly instead? Something like this?
```
from numpy.fft import _pocketfft as npfft
```
and use npfft.* for the reference calls

File: mkl_fft/_fft_utils.py [hunk 6]
```diff
@@ -78,6 +78,37 @@ def _compute_fwd_scale(norm, n, shape):
         return np.sqrt(fsc)
 
 
+def _compute_nd_scale_shape(x, s, axes, norm=None, invreal=False):
```
vlad-perevezentsev: This function almost exactly replicates the logic in `_cook_nd_args`
It looks like we can reuse it
Something like this
```
def _nd_fwd_scale(x, s, axes, norm, invreal=False):
    """Forward scale of an N-D transform, normalized over the transformed axes.

    ``_compute_fwd_scale`` falls back to the full array shape when ``s`` is
    None, which over-normalizes a subset-of-axes transform; for c2r the basis
    is the output length ``2 * (n - 1)``. Both are exactly what
    ``_cook_nd_args`` already resolves for the transform itself.
    """
    _check_norm(norm)
    if norm in (None, "backward"):
        return 1.0
    if s is None:
        s, _ = _cook_nd_args(x, s, axes, invreal=invreal)
    return _compute_fwd_scale(norm, s, x.shape)
```

`return _compute_fwd_scale(norm, s, x.shape)` will also help reduce the code in the fft functions 
```
fsc = _nd_fwd_scale(x, s, axes, norm)

#instead of 

ss = _compute_nd_scale_shape(x, s, axes, norm)
fsc = _compute_fwd_scale(norm, ss, x.shape)
```

CHANGELOG.md
```diff
@@ -14,6 +14,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Pinned Cython in the Coverity Scan workflow so generated code stays stable between scans, and added `coverity/README.md` documenting the known false-positive families and the scan review checklist [gh-374](https://github.com/IntelPython/mkl_fft/pull/374)
 
 ### Fixed
+* Fixed `norm="forward"`/`"ortho"` scaling in `fftn`, `ifftn`, `rfftn`, `irfftn` and the `fft2` family when only a subset of axes is transformed: the scale used the full array shape instead of the transformed axes. The `numpy_fft` and `scipy_fft` interfaces were unaffected [gh-370](https://github.com/IntelPython/mkl_fft/pull/370)
+* Fixed `norm="forward"`/`"ortho"` scaling in `irfftn` and `irfft2`, which normalized over the input length `n` rather than the complex-to-real output length `2 * (n - 1)` [gh-370](https://github.com/IntelPython/mkl_fft/pull/370)
```
vlad-perevezentsev, line 18: The changelog does not mention gh-336 which this PR also fixes

mkl_fft/tests/test_dispatch_equivalence.py
```diff
@@ -0,0 +1,422 @@
...
+    x = _make(_SHAPE_3D, dtype)
+    got = getattr(mkl_fft, func)(x, axes=axes, norm=norm)
+    want = getattr(np.fft, func)(x, axes=axes, norm=norm)
+    _check(got, want, dtype)
+
+
+@pytest.mark.parametrize("func", ["fftn", "ifftn", "fft2", "ifft2"])
+@pytest.mark.parametrize("dtype", ["float64", "complex128"])
+@pytest.mark.parametrize("norm", [None, "backward", "forward", "ortho"])
+def test_empty_axes_is_identity(func, dtype, norm):
```
vlad-perevezentsev: It overlaps `test_empty_axes` and `test_empty_axes_returns_same_object` in `test_fftnd.py` almost entirely 
Could be removed

mkl_fft/tests/test_dispatch_equivalence.py
```diff
@@ -0,0 +1,422 @@
...
+    _check(got, want, dtype)
+
+
+@pytest.mark.parametrize("func", ["rfftn", "irfftn"])
+@pytest.mark.parametrize("dtype", _DTYPES)
+@pytest.mark.parametrize("axes", _AXES_3D)
+@pytest.mark.parametrize("order", _ORDERS)
+def test_rfftn_axes_subset(func, dtype, axes, order):
+    if func == "rfftn" and dtype not in _REAL_DTYPES:
+        pytest.skip("rfftn takes real input")
```
vlad-perevezentsev: You can parameterize it here and below to avoid skipping lines in the test log
```
_R2C_CASES = [("rfftn", dt) for dt in _REAL_DTYPES] + [
    ("irfftn", dt) for dt in _DTYPES
]

@pytest.mark.parametrize("func,dtype", _R2C_CASES)
```

mkl_fft/tests/test_dispatch_equivalence.py
```diff
@@ -0,0 +1,422 @@
...
+# random data over these lengths stay well inside 2e-5.
+_TOL = {
+    "float32": 2e-5,
+    "complex64": 2e-5,
+    "float64": 1e-12,
+    "complex128": 1e-12,
+}
+
+# every non-empty subset of the axes of a 3-D array, plus None
+_AXES_3D = [
```
vlad-perevezentsev: Could we add negative axes to `_AXES_3D`?

mkl_fft/tests/test_dispatch_equivalence.py
```diff
@@ -0,0 +1,422 @@
...
+    ignores ``s`` and falls back to the input-derived basis would still pass
+    that one case -- 20 and 10 are what catch it. A regression that instead
+    doubles ``s[-1]`` itself is caught by all three.
+    """
+    x = _make(_SHAPE_3D, "complex128")
+    got = mkl_fft.irfftn(x, s=s, axes=(0, 1, 2), norm=norm)
+    want = np.fft.irfftn(x, s=s, axes=(0, 1, 2), norm=norm)
+    _check(got, want, "complex128")
+
+
```
vlad-perevezentsev, line 425: A few combinations on the new code path are not covered yet : negative axes with norm . irfftn with subset axes, s, and norm , out= with s= and norm and fft2/rfft2/irfft2 with s and norm

mkl_fft/tests/test_dispatch_equivalence.py
jharlow-intel: Added a check to account for the numpy fft interface being non-standard during testing, good catch!

mkl_fft/_fft_utils.py
jharlow-intel: added a new function `_nd_fwd_scale` that re-uses cook_nd_args as you described, thanks!

jharlow-intel:
@vlad-perevezentsev  review changes implemented

ndgrigorian:
@jharlow-intel please fix the pre-commit in this PR, I think it's more or less ready to go in and we can knock out the pre-commit update fix here

jharlow-intel:
@ndgrigorian  fixed 👍

ndgrigorian (approved):
In general this LGTM, thank you @jharlow-intel

PR #375, opened 2026-09-03
ci: harden GitHub Actions workflows and add zizmor scan

antonwolfy:
Backport of IntelPython/dpctl#2374, adapted to mkl_fft's own set of workflows.

Adds a CI job that runs the [zizmor](https://docs.zizmor.sh) static analyzer over the workflow files under `.github/`. `zizmor` audits GitHub Actions workflows for supply-chain and privilege-escalation weaknesses — credential persistence through the checkout token, overly broad `GITHUB_TOKEN` permissions, unpinned action references, and template injection via `${{ ... }}` expansion in `run:` blocks.

Alongside the new scan, this applies the corresponding hardening to the existing workflows so they pass the audit:

- add `persist-credentials: false` to every `actions/checkout` step that lacked it
- narrow top-level `permissions: read-all` to `permissions: contents: read`
- add a 7-day `cooldown` to the dependabot update entries

The dpctl-only changes do not apply here: mkl_fft has no `backfill-docs.yml` (nor any other `${{ github.event.inputs.* }}` used inside a `run:` block), and it does not use `mshick/add-pr-comment`, so the malformed version-pin fix is not relevant.

This is a CI/configuration-only change; no library code, tests, or documentation are affected.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,6 +5,8 @@ updates:
     schedule:
       interval: "weekly"
       day: "saturday"
+    cooldown:
+      default-days: 7
     rebase-strategy: "disabled"
     labels: ["autoupdate"]
     groups:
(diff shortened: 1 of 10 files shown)


vlad-perevezentsev (approved):
LGTM
Thank you @antonwolfy

